PR #10, opened 2015-12-02
added logger for records missing countries

### LINKED ISSUE #11: Add exception logger to WOF importer
One criteria we're using for WOF is that a record has to have a country to be imported.  Add a logger that warns when a record can't be walked up to a country by parent_id.

stephenkhess:
also fixed bug where `placetype` was being used instead of `place_type`

Fixes #11

### DIFF
--- a/src/importStream.js
+++ b/src/importStream.js
@@ -1,53 +1,80 @@
 var event_stream = require('event-stream');
 var map_stream = require('through2-map');
+var spy_stream = require('through2-spy');
+var _ = require('lodash');
 
 var Document = require('pelias-model').Document;
+var peliasLogger = require( 'pelias-logger' );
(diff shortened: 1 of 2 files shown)


src/importStream.js
```diff
@@ -1,53 +1,80 @@
...
+    ]
+  });
+
   var object_getter_stream = map_stream.obj(function(id) {
     return records[id];
   });
 
-  var document_stream = map_stream.obj(function(record) {
-    // console.log('starting ' + record.id + ': ' + record.name);
+  var has_country_validation_stream = spy_stream.obj(function(wofDoc) {
```
orangejulius: nice use of spy-stream!

orangejulius:
Nice :+1:

stephenkhess:
Merged

PR #12, opened 2015-12-04
Add unit tests

stephenkhess:
readStreamTest.js test is a bit tedious but setup wasn't all that bad.  Alternatively the strategy would be to inject either an array of streams or an options object containing all the requested streams.  I'd prefer the former but I couldn't figure out how to iterate a list and chain together the pipes.

### DIFF
--- a/test/readStreamTest.js
+++ b/test/readStreamTest.js
@@ -4,27 +4,10 @@ var fs = require('fs-extra');
 
 var readStream = require('../src/readStream');
 
-/*
- * Test a stream with the following process:
- * 1. take input as array
- * 2. turn the array into a stream
- * 3. pipe through stream to be tested
- * 4. pass output of that stream to a callback function
- * 5. assertions can then be made in that callback
- *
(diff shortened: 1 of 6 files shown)


stephenkhess:
Fixed #12

test/readStreamTest.js
```diff
@@ -0,0 +1,93 @@
...
+var readStream = require('../src/readStream');
+
+tape('readStream', function(test) {
+  /*
+    this test is not terribly attractive, i'm not happy with it but setup wasn't
+    all that painful.
+  */
+  test.test('readStream should return from all requested types and populate wofRecords', function(t) {
+    // remove tmp directory if for some reason it's been hanging around from a previous run
+    fs.removeSync('tmp');
```
dianashk: I would pull all the file writing and dir creating into a setup function so it's more clear where the test actually begins.

stephenkhess: Good call, will do

orangejulius:
:+1: :+1:

PR #17, opened 2015-12-21
Set alpha3

### LINKED ISSUE #16: Added ISO3 suppport
pull from WOF record, but only for records that have a country

the-saltiest-dog:
WOF only exports iso2 currently, so I included support for iso3 using the iso3166-1 node module

fixes #16

### DIFF
--- a/src/hierarchyFinder.js
+++ b/src/hierarchyFinder.js
@@ -1,0 +1,62 @@
+var _ = require('lodash');
+
+module.exports = {};
+
+var hasName = function(r) {
+  return r.name;
+};
+
+var isDefined = function(r) {
+  return r;
+};
+
+/*
+  This function builds a hierarchy by starting with the current record and
+  walking up by parent_id until a parent can't be found, filtering out those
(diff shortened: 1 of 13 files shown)


src/hierarchyFinder.js
```diff
@@ -0,0 +1,62 @@
+var _ = require('lodash');
+
+module.exports = {};
+
+var hasName = function(r) {
+  return r.name;
```
orangejulius: Would we ever want a record where the name is the empty string (`""`) to pass this function? The number zero, `false` and the empty string will all evaluate to false in Javascript conditionals.

src/hierarchyFinder.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+var isDefined = function(r) {
```
orangejulius: What about removing this function and using _.isObject? I think all the records coming through should be objects, and that way we can use a built in Lodash function.

orangejulius:
:+1:

PR #23, opened 2016-01-22
Support new pelias model

trescube:
adds:

- configuration file support
- `npm run import`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Importer for Who's on First",
   "main": "index.js",
   "scripts": {
+    "import": "node import.js",
     "test": "node test/test | tap-dot"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Importer for Who's on First",
   "main": "index.js",
   "scripts": {
-    "import": "node import.js",
+    "start": "node import.js",
     "test": "node test/test | tap-dot"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "pelias-config": "^1.0.3",
     "pelias-dbclient": "0.0.9",
     "pelias-logger": "0.0.8",
-    "pelias-model": "pelias/model#c97cf830d4fe9cb8333f714ee70a72b476779e9c",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Importer for Who's on First",
   "main": "index.js",
   "scripts": {
+    "start": "node import.js",
```
orangejulius: nice!

src/peliasDocGenerators.js
```diff
@@ -42,21 +43,28 @@ module.exports.createPeliasDocGenerator = function(hierarchy_finder) {
...
+          wofDoc.addParent('localadmin', hierarchy_element.name, hierarchy_element.id.toString());
           break;
         case 'county':
           wofDoc.setAdmin( 'admin2', hierarchy_element.name);
+          wofDoc.addParent('county', hierarchy_element.name, hierarchy_element.id.toString());
           break;
         case 'region':
           wofDoc.setAdmin( 'admin1', hierarchy_element.name);
-          if (hierarchy_element.hasOwnProperty('abbreviation')) {
+          if (!_.isUndefined(hierarchy_element.abbreviation)) {
```
orangejulius: Assuming valid `heirarchy_element.abbreviation` values don't include `null`, the integer `0`, `false`, etc, I think this can be simplified to just `if (heirarchy_element.abbreviation)`

test/readStreamTest.js
```diff
@@ -9,6 +9,70 @@ tape('readStream', function(test) {
     all that painful.
   */
   test.test('readStream should return from all requested types and populate wofRecords', function(t) {
+    function setupTestEnvironment() {
```
orangejulius: Not required, but i found the [`temp` library](https://www.npmjs.com/package/temp) that can handle temporary files in a platform agnostic way, and automatically clean up. See [https://github.com/pelias/openaddresses/blob/master/test/parameters.js](https://github.com/pelias/openaddresses/blob/master/test/parameters.js) for usage

orangejulius:
:+1:

dianashk:
@trescube it's strange that the travis test fails for node 0.12 but works fine for 0.10

PR #36, opened 2016-02-05
Improve travis

orangejulius:
Makes the TravisCI configuration more similar to the best practices (connects pelias/pelias#258)

Some differences:
* There is no functional test yet
* Node 0.10 is no longer supported here, so allow failures with that version.


dianashk:
:100:

PR #46, opened 2016-02-26
45 use iso2 from hierarchy for alpha3 lookup

### LINKED ISSUE #45: switch to hierarchy for alpha3 lookup
ISO2 isn't reliable for placetype < country, so work up thru hierarchy to placetype==country record, then use that ISO2 for records' alpha3

trescube:
Fixed #45


dianashk:
:rocket:

PR #60, opened 2016-03-31
Add `zs:pop10` value for population candidate value

### LINKED ISSUE #59: Add population support from `zs:pop10`
Neighborhoods (especially in NYC) lack a `gn:population` field which is causing adverse effects since our scoring logic punishes results with popularity but no population (eg - Chelsea).  The whosonfirst importer should use the `zs:pop10` value for population when `gn:population` doesn't exist.  Also, of the 43,571 WOF records containing `zs:pop10`, 1,670 have a `0` value.  In cases where `zs:pop10` has a `0` value, it should be ignored so that our scoring logic can give it a default of `1`.

trescube:
NYC (and other neighborhoods) don't appear to have `gn:population` values, so inputs like `Chelsea` are being punished in scoring for no other reason than lack of value (even the default `1` value isn't helping).  This fix uses the `zs:pop10` value for population when non-zero and `gn:population` is not available.  It's likely that `gn:population` and `zs:pop10` presence are mutually exclusive.

Fixed #59
Fixed pelias/pelias#300

### DIFF
--- a/src/readStreamComponents.js
+++ b/src/readStreamComponents.js
@@ -81,6 +81,17 @@ function isUsCounty(base_record, qs_a2_alt) {
           !_.isUndefined(qs_a2_alt);
 }
 
+// this function favors gn:population when available, falling back to zs:pop10
+//  when available and > 0
+function getPopulation(properties) {
+  if (properties['gn:population']) {
+    return properties['gn:population'];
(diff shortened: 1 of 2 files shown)


orangejulius:
I think the test results look good, and in principle this is a strictly excellent change anyway. :+1:

PR #62, opened 2016-03-31
Simply read stream and componentize

### LINKED ISSUE #61: Refactor whosonfirst importer
All the streams are part of 1 big library which makes tests unwieldy and hard to read.  Additionally, the code relies upon the data file path being in the meta file, which takes some sanitizing.  Refactor to generate filename from id field in metadata.

trescube:
The code suffers from 2 problems:

- most stream components are in 1 big file which is unwieldy, hard to read, and poorly named
- data file loading is performed using the file path value from the metadata which needs to be normalized

This PR solves both problems:

- extracts out the individual streams to separate files with better names
- calculates the data file name from the id which has a consistent convention

Fixed #61


orangejulius:
Nice! :+1:

orangejulius:
Since we've now done basically this same thing in all our importers, we should start to look and see if any of the parts of the import pipeline can be extracted into separate modules.

dianashk:
Woah! This is great!

src/components/calculateFilePath.js
```diff
@@ -0,0 +1,16 @@
+var map = require('through2-map');
+var sep = require('path').sep;
+
+module.exports.create = function() {
```
orangejulius: give this function a name so our stack traces aren't anonymously sad

src/components/fileIsReadable.js
```diff
@@ -0,0 +1,19 @@
+var filter = require('through2-filter');
+var fs = require('fs');
+var path = require('path');
+var util = require('util');
+
+module.exports.create = function(dataDirectory) {
```
orangejulius: Also needs a name

src/components/isValidId.js
```diff
@@ -0,0 +1,9 @@
+var filter = require('through2-filter');
+
+module.exports.create = function() {
```
orangejulius: Even 5 line functions deserve a name!

src/components/recordHasIdAndProperties.js
```diff
@@ -0,0 +1,10 @@
+var filter = require('through2-filter');
+
+/*
+  This function filters out incomplete records
+*/
+module.exports.create = function() {
```
orangejulius: And yes even 3 line functions!

trescube:
I'm doing some research to see how much simple it now is to just use the `path` field is.  It used to be somewhat unreliable, it may be better now.

trescube:
The `path` field is much more reliable now so I'll remove calculating path from id.

orangejulius:
Looks good! :+1:

PR #64, opened 2016-04-04
Add borough support

### LINKED ISSUE #63: add `borough` place type support
Now that WOF has added support for NYC boroughs, we need to import them.

trescube:
Fixed #63


orangejulius:
So does this include #62 ?

trescube:
I branched from #62

orangejulius:
Cool. How about we merge 62 after the last couple changes that I mentioned, then this will show just the actual changes needed for borough in the diff?

trescube:
Sounds good

PR #69, opened 2016-04-05
Add borough hierarchy element to Pelias docs

### LINKED ISSUE #68: add `borough` hierarchy element support
Documents added in WOF can be children of `borough`s so add into hierarchy.

trescube:
Fixed #68


orangejulius:
:+1:

PR #71, opened 2016-04-05
added helper methods and check for `mz:is_current` === 0 (inactive)

### LINKED ISSUE #70: Add additional active record checks
WOF considers records with edtf:superseded (any value) or mz:is_current (0) properties to be inactive, at least not for Pelias consumption.  Add these checks to avoid loading them.

trescube:
Support edtf:superseded and mz:is_current checks for active records. Fixed #70


orangejulius:
:+1:

PR #86, opened 2016-04-22
Fix model usage

orangejulius:
We were bitten by the same test suite bug as pelias/pelias#500 where failures are not properly recorded, and let #80 merge when it broke things. Here's the test output on travis

https://travis-ci.org/pelias/whosonfirst/jobs/124026505

This removes all usage of `setAdmin`/`getAdmin` which is now deprecated since it only updates the old admin fields. All the hierarchy fields were already being properly set.

One other thing I discovered: there is no analog to `getAdmin` with our new admin heirarchy in pelias-model, so the stream that logs when a record has no country had to go away for now.

### DIFF
--- a/src/importStream.js
+++ b/src/importStream.js
@@ -1,8 +1,4 @@
-var spy_stream = require('through2-spy');
-var _ = require('lodash');
-
 var peliasModel = require( 'pelias-model' );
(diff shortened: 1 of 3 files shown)


trescube:
👍 thx for getting rid of the `has_country_validation_stream` stuff, that was very old and ready for removal

PR #95, opened 2016-06-08
Use label centroid over math centroid if it exists

orangejulius:
I'm not sure we want to do this in 100% of cases, but we want it in at
least most cases.

Fixes https://github.com/pelias/pelias/issues/356


orangejulius:
Merging based on feedback from standup :)

PR #98, opened 2016-06-13
Check node version

orangejulius:
Use [node-version-checker](https://www.npmjs.com/package/node-version-checker) to ensure Node 0.12 or above is used. Node 0.10 is not supported and fails in a confusing way. Thanks @easherma for helping us notice!

It generates a _reasonably_ friendly error message:

```Error: Current node version does not match the engines section of the package.json. Wanted >=0.12 but was v0.10.43```


trescube:
👍

PR #100, opened 2016-06-13
fix: Use regex to test for correct error

orangejulius:
In Node 6, the error output for JSON parse exceptions now includes a
nice backtrace. Unfortunately our test was comparing exact output, so it
now fails. This regex checks that the part we care about is there
without expecting a specific output.


trescube:
👍

orangejulius:
For some reason semantic-release is not properly ignoring that the node 6 build fails, so until we merge this we don't get any new NPM packages :(

Issue #106, opened 2016-06-30
import population data from 'qs:pop'


missinglink:
the following record has a property called `qs:pop` which contains population info, that property may be new and is not currently being checked during import.

- https://whosonfirst.mapzen.com/spelunker/id/404484357/#12/40.6178/-80.2627
- https://github.com/pelias/whosonfirst/blob/master/src/components/extractFields.js#L13-L19

orangejulius:
Sounds good to me. @nvkelso, is this a field we should be using? 

Right now we first look at `gn:population` and then `zs:pop10`. Where should `qs:pop` go in that ordering?

trescube:
Or have a general `mz:population` field would be ideal so we don't have to look in one place for neighbourhoods and another for localities.

nvkelso:
`mz:population` is my preference

orangejulius:
Awesome, I suspect we'll want to handle `qs:pop` ourselves in the meantime (we don't want to rush you to release it when the change on our side is relatively small). Any suggestion for an ordering of the 3 variables?

trescube:
I believe that `zs:pop10` only applies to neighbourhoods.

dianashk:
Update the code to import all 3: `gn:population`, `qs:pop`, `zs:pop10`.
Will switch to `mz:population` when that becomes available.

PR #116, opened 2016-07-27
Added qs:pop and mz:population options for population

avulfson17:
Fixes pelias/whosonfirst#106


src/components/extractFields.js
```diff
@@ -10,11 +10,16 @@ function isUsCounty(base_record, qs_a2_alt) {
 
 // this function favors gn:population when available, falling back to zs:pop10
 //  when available and > 0
+// the mz:population should be moved to the top when implemented
```
trescube: If `mz:population` is moved to the top now then we won't have to change the code when it's supported.

trescube:
:shipit:

orangejulius:
LGTM

Issue #94, opened 2016-06-03
Import Who's on First venues


orangejulius:
Who's on First now includes many venues. The data is split across several hundred repos in the [whosonfirst-data](https://github.com/whosonfirst-data/) Github organization, so a big challenge will simply be gathering all the data. Several of the repositories use git-lfs as well. 

On the importer side, we are currently able to squeeze all the WOF administrative area records into memory, which obviously won't work with millions of venues.
## has to be done to allow for dev work
- [x] ~~create script to set up test data directory with example venue data~~ (no longer needed, because the venue bundles are published and can be downloaded directly)
- [x] publish WIP code to support better memory management in importer (https://github.com/pelias/whosonfirst/pull/119, related to https://github.com/pelias/whosonfirst/issues/7)

## has to be done before production readiness
- [x] improve WOF venue generation code to build more venue bundles (preferably we get one giant venue bundle) (this has been done by the Who's on First team!)
- [x] change list of WOF layers in API so venue and address records in WOF can be queried (similar to https://github.com/pelias/api/issues/569) (fixed in https://github.com/pelias/api/pull/645)
- [x] ensure admin lookup code doesn't try to load all the venues if it's pointed at a directory containging WOF admin and venue data (update: it doesn't, because like the wof importer, wof-pip-service [uses the meta files](https://github.com/pelias/wof-pip-service/blob/master/src/index.js#L25-L36) to know what to load)
- [x] (Mapzen Search specific) update chef scripts to download venue bundles OR update chef scripts to use Javascript downloader, and update that to include venue data (preferred). For now, this can just be a big hardcoded list
- [x] Update whosonfirst repo readme with venue information
- [x] Introduce configuration option to disable importing venues (https://github.com/pelias/whosonfirst/pull/142)
- [ ] review any acceptance test failures with a full planet build and resolve them to our satisfaction
- [ ] Ensure performance is reasonable, since this may add 10-15 million new records!
- [ ] Update the [installation docs](https://github.com/pelias/pelias-doc/blob/master/installing.md) and [data sources docs](https://github.com/pelias/pelias-doc/blob/master/data-sources.md)

## can be done as follow up improvements
- [x] Rewrite downloader script (again) to be faster and smarter about downloading venue data (https://github.com/pelias/whosonfirst/issues/135)
- [ ] write code that allows us to handle street addresses in WOF records. this can mirror the address duplicating code in the OSM importer
[...]

PR #119, opened 2016-08-03
Use a single stream for importing records

### LINKED ISSUE #7: Store fewer objects in memory during import
Right now this importer stores an object for every single WOF item in memory. While that object doesn't have all the fields from the WOF item in it (polygon data, most notably, isn't stored), it still takes up quite a bit of memory.

As WOF grows or this importer loads more fields, it will almost certainly brush up closer against the Node 1GB memory limit. It's probably close enough to it now that performance is suffering.

One solution would be to store only admin areas, to speed up populating the heirarchy. This would be ideal if it can safely fit in memory.

 The other option is to not store anything, and re-load admin areas each time we are populating an item. The operating system's file cache would speed this up partially, but re-parsing of all the JSON would still have to be done, which is slow.

### LINKED ISSUE #101: Show progress when initially loading data
The importer is currently very quiet for the first few minutes when being run. In the background it is (hopefully) loading all the WOF admin data into memory, but since there's no way to judge progress, there's no way to know its working. If you aren't expecting this, it's especially concerning. Perhaps as part of #7 we should improve this.

The download script is similarly quiet.

Thanks to @easherma for bringing this to our attention.

orangejulius:
Previously, the WOF importer loaded all records into memory in one
stream, and then processed and indexed the records in Elasticsearch in a
second stream after the first stream was done.

This has several problems:
* It requires that all data can fit into memory. While this is not
  _so_ bad for WOF admin data, where a reasonably new machine can handle
  things just fine, it's horrible for venue data, where there are already
  10s of millions of records and will likely be many more in the future.
* Its slower: by separating the disk and network I/O sections, they
  can't be interleaved to speed things up.
* It doesn't give good feedback when running the importer that something
  is happening: the importer sits for several minutes loading records
  before the dbclient progress logs start displaying

This change fixes all those issues, by processing all records in a
single stream, starting at the highest hierarchy level, and finishing at
the lowest, so that all records always have the admin data they need to
be processed.

A change like this is necessary to support Who's on First venues, and in fact this code has already been tested by importing about 1M venues from California!

Fixes #101
Connects #7 (it doesn't quite fix it, for that we need to be able to not even store all _admin_ areas at once, for example to import geometries)
Connects #94

### DIFF
--- a/import.js
+++ b/import.js
@@ -1,9 +1,8 @@
 var peliasConfig = require( 'pelias-config' ).generate();
-var readStream = require('./src/readStream');
+var readStreamModule = require('./src/readStream');
 var importStream = require('./src/importStream');
 var peliasDbclient = require( 'pelias-dbclient' );
 var peliasDocGenerators = require('./src/peliasDocGenerators');
-var wofRecordStream = require('./src/wofRecordStream');
 var hierarchyFinder = require('./src/hierarchyFinder');
 var checker = require('node-version-checker').default;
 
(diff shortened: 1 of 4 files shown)


import.js
```diff
@@ -25,39 +24,38 @@ if (directory.slice(-1) !== '/') {
...
-  // a stream of WOF records
-  var recordStream = wofRecordStream.createWofRecordsStream(wofRecords);
+// a cache of only admin records, to be used to fill the hierarchy
+// of other, lower admin records as well as venues
+var wofAdminRecords = {};
 
-  // how to convert WOF records to Pelias Documents
-  var documentGenerator = peliasDocGenerators.create(
-    hierarchyFinder.hierarchies_walker(wofRecords));
+var readStream = readStream.create(directory, types, wofAdminRecords);
```
dianashk: It's a bit confusing that you redefine `readStream` here and assign `readStream.create()` to it. Would be great if this variable or the one in the require block had a different name.

dianashk:
Other than the one confusing variable name, code looks solid.

import.js
orangejulius: oh, doh! good catch

orangejulius:
Variable name is fixed!
I also did some timing of this branch vs master. With a freshtly emptied and restarted Elasticsearch each time, master runs in `9m35s`, and this branch runs in `6m37s`! Also worth noting there are now 420k admin records in Who's on First!

dianashk:
:shipit:

missinglink:
:shipit:

PR #122, opened 2016-08-05
Use lbl:bbox property when present

orangejulius:
After the discussion in https://github.com/pelias/pelias/issues/370,
Who's on First implemented the `lbl:bbox` field, which is a hand-made
bounding box for areas where the mathematically derived bounding box
doesn't match up with what people generally want.

At least two of these were implemented in
https://github.com/whosonfirst-data/whosonfirst-data/issues/361, so it's
time to start using them!

Connects https://github.com/pelias/pelias/issues/370
Connects https://github.com/pelias/pelias/issues/397

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -38,6 +38,14 @@ function getLon(properties) {
   }
 }
 
+function getBoundingBox(properties) {
+  if (properties['lbl:bbox']) {
+    return properties['lbl:bbox'];
+  } else {
+    return properties['geom:bbox'];
+  }
+}
+
 /*
   This function extracts the fields from the json_object that we're interested
   in for creating Pelias Document objects.  If there is no hierarchy then a
@@ -54,7 +62,7 @@ module.exports.create = function map_fields_stream() {
       parent_id: json_object.properties['wof:parent_id'],
       lat: getLat(json_object.properties),
       lon: getLon(json_object.properties),
-      bounding_box: json_object.properties['geom:bbox'],
+      bounding_box: getBoundingBox(json_object.properties),
       iso2: json_object.properties['iso:country'],
       population: getPopulation(json_object.properties),
       popularity: json_object.properties['misc:photo_sum']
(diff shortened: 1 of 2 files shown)


src/components/extractFields.js
```diff
@@ -38,6 +38,14 @@ function getLon(properties) {
   }
 }
 
+function getBoundingBox(properties) {
+  if (properties['lbl:bbox']) {
```
dianashk: would be better to use hasOwnProperty or isSet here

orangejulius: Hm I think you're right, as well as for the other helper functions. For lat/lon for example, I bet we don't handle null island very well.

orangejulius: Upon further reflection, I now disagree (although not very strongly). The only differences between `if (object['prop'))` and `if (obect.hasOwnProperty('prop'))` are how it handles the number zero and empty strings. For bounding boxes, neither of those are valid values, so falling back to the standard bbox is fine.

For other fields it's mostly the same way. With population we explicitly want to ignore zero values and fall back, and while I mentioned the null island case (where we actually _do_ [incorrectly fall back to the `geom` lat/lon because the `lbl` lat/lon is 0, 0](https://whosonfirst.mapzen.com/spelunker/id/1/#18/-0.00004/-0.00013), it feels like the simplicity of the current syntax is worth the slight mishandling of that one case.

Am I missing an edge case that we are protected against with `hasOwnProperty`? I think I covered [all of them](https://javascriptweblog.wordpress.com/2011/02/07/truth-equality-and-javascript/)

dianashk: I think you're right about the various conditions and the code will work fine for this scenario. However, it's not a great convention and doesn't make the true intent of this statement clear. I recall discussing this with the team a long time ago and coming to an agreement to use hasOwnProperty or isSet checks for these cases. It's not a big deal, but I think it makes the code a bit easier to understand.

orangejulius: You're right, i remember that conversation as well. Code updated!

dianashk:
LGTM 🌴 🌴 🌴

trescube:
:shipit:

PR #125, opened 2016-08-08
Add dependency support

trescube:
Without `dependency` support, cities and regions won't resolve since libpostal treats dependencies as countries.  For example, "San Juan, Puerto Rico".

### DIFF
--- a/test/peliasDocGeneratorsTest.js
+++ b/test/peliasDocGeneratorsTest.js
@@ -27,7 +27,7 @@ tape('create', function(test) {
         name: 'name 2',
         lat: 13.131313,
         lon: 31.313131,
-        place_type: 'macroregion',
+        place_type: 'dependency',
         bounding_box: '-13.691314,49.909613,1.771169,60.847887'
       },
       3: {
@@ -35,7 +35,7 @@ tape('create', function(test) {
         name: 'name 3',
         lat: 14.141414,
         lon: 41.414141,
-        place_type: 'region',
+        place_type: 'macroregion',
         bounding_box: '-13.691314,49.909613,1.771169,60.847888'
       },
       4: {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "pelias-config": "2.1.0",
     "pelias-dbclient": "2.0.0",
     "pelias-logger": "0.0.8",
-    "pelias-model": "4.1.0",
+    "pelias-model": "4.2.0",
```
orangejulius: The upgrade to 4.2.0 was [already merged](https://github.com/pelias/whosonfirst/pull/124). In general Greenkeeper always handles these quite quickly :)

orangejulius:
LGTM

package.json
trescube: Smooth

test/peliasDocGeneratorsTest.js
```diff
@@ -27,84 +27,94 @@ tape('create', function(test) {
...
+        bounding_box: '-13.691314,49.909613,1.771169,60.847893'
+      },
+      9: {
+        id: 9,
+        name: 'name 9',
+        lat: 10.101010,
+        lon: 1.010101,
         place_type: 'locality',
-        bounding_box: '-13.691314,49.909613,1.771169,60.847893',
+        bounding_box: '-13.691314,49.909613,1.771169,60.847894',
```
orangejulius: This test, as evidenced by the weird diff in the many lines above, is getting unweildy. Maybe we can split it up. Do all these cases really have to be tested in a stream?

orangejulius: (we can and should address this in a different PR)

dianashk:
:shipit:

PR #126, opened 2016-08-09
added `dependency` abbreviation support

trescube:
This will allow us to match the `PR` in `San Juan, PR`.  Will not work immediately, waiting on https://github.com/whosonfirst-data/whosonfirst-data/issues/439.

### DIFF
--- a/src/peliasDocGenerators.js
+++ b/src/peliasDocGenerators.js
@@ -47,6 +47,9 @@ module.exports.create = function(hierarchy_finder) {
     // iterate the hierarchy, assigning fields
     hierarchy_finder(record).forEach(function(hierarchy_element) {
       switch (hierarchy_element.place_type) {
+        case 'neighbourhood':
+          wofDoc.addParent('neighbourhood', hierarchy_element.name, hierarchy_element.id.toString());
+          break;
         case 'locality':
           wofDoc.addParent('locality', hierarchy_element.name, hierarchy_element.id.toString());
           break;
--- a/src/peliasDocGenerators.js
+++ b/src/peliasDocGenerators.js
@@ -63,7 +63,7 @@ module.exports.create = function(hierarchy_finder) {
           wofDoc.addParent('macrocounty', hierarchy_element.name, hierarchy_element.id.toString());
           break;
         case 'region':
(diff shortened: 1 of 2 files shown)


src/peliasDocGenerators.js
```diff
@@ -73,7 +73,11 @@ module.exports.create = function(hierarchy_finder) {
           wofDoc.addParent('macroregion', hierarchy_element.name, hierarchy_element.id.toString());
           break;
         case 'dependency':
-          wofDoc.addParent('dependency', hierarchy_element.name, hierarchy_element.id.toString());
+          if (hierarchy_element.abbreviation) {
```
orangejulius: Someday we will refactor this function. Maybe it will take an improvement to the model code? (it's fine for now though)

trescube: It will take improvements to the model code because it blows up gloriously on empty/null values due to the validation routines.

orangejulius:
LGTM

src/peliasDocGenerators.js
dianashk: let's use the `hasOwnProperty` or the [lodash `has` function](https://lodash.com/docs#has) here to check for abbreviation.

trescube:
Apparently we weren't including `neighbourhood`'s in parentage either so this includes as well.

src/peliasDocGenerators.js
```diff
@@ -47,6 +47,9 @@ module.exports.create = function(hierarchy_finder) {
     // iterate the hierarchy, assigning fields
     hierarchy_finder(record).forEach(function(hierarchy_element) {
       switch (hierarchy_element.place_type) {
+        case 'neighbourhood':
+          wofDoc.addParent('neighbourhood', hierarchy_element.name, hierarchy_element.id.toString());
+          break;
```
orangejulius: Wait so we haven't had neighbourhoods in the parent heirarchy at all? Oh man..

trescube: Correct.  It probably hasn't been a big deal yet but will be for libpostal integration.

dianashk:
:shipit:

PR #136, opened 2016-09-14
New bundle downloader with venue support

### LINKED ISSUE #135: Re-rewrite downloader
With [WOF veneus](https://github.com/pelias/whosonfirst/issues/94) comming soon, we will often be downloading significantly more data, so the downloader script needs to be made faster. This means doing the following:

- [ ] **Using native command line tools like wget and tar**. We've chatted and are making the decision to drop any support for Windows that was around because we had a pure Node.js solution. It may not happen right away but we can add that back later with Cygwin, etc. The command line tools are just [too fast](https://github.com/pelias/whosonfirst/pull/111#issuecomment-232227743) and memory efficient compared to Node.js to not use.
- [ ] **Download and extract multiple bundles simultaneously**. Networks and CPUs will be most effectively utilized with multiple downloads happening at once. This will require a bit more coordination for error handling, but is worth it
- [ ] **Calculates all the possible WOF venue bundles from an authoritative source** The file at https://github.com/whosonfirst-data/whosonfirst-data-venue/blob/master/data.json lists all the Who's on First venue repos, which can be used to derive the name of all the bundle tar files to download. There are lots of them, and they will change over time, so its not good in the long term to hardcode them.

orangejulius:
This is a new bundle downloader that supports venues.  It uses shell commands and parallelization to be much faster than the previous one, with little risk of running out of memory.

The readme is also updated, with sections reordered so that a user following from top to bottom will read the things they need in the order they need them.

Please test out the downloader with at least `npm run download -- --admin-only` or, if you are ambitious, patient, and have lots of free disk space, `npm run download`.

Fixes #135


README.md
```diff
@@ -43,31 +40,35 @@ Other types may be included in the future.
...
-## Data
+## Configuration
+
+The sole Pelias configuration option available is `imports.whosonfirst.datapath`.  It should point to the root of where Who's On First data has been downloaded.
 
-There are multiple ways to download Who's On First data
+## Downloading the Data
 
-* The enclosed script `download_data.js` will download the required bundles and place the data into `./wof_data` in the required directory layout.  To install the required node module dependencies and run the download script:
+* The enclosed script `download_data.js` will download the required bundles and place the data into the datapath configured in [pelias-config](https://github.com/pelias/config) in the required directory layout.  To install the required node module dependencies and run the download script:
```
trescube: Nice!  Glad it goes to a configurable directory now.

src/bundleList.js
```diff
@@ -0,0 +1,302 @@
...
+  'venue-us-sd',
+  'venue-us-vt',
+  'venue-us-nj',
+  'venue-us-wv',
+  'venue-us-il',
+  'venue-us-or',
+  'venue-us-wy',
+  'venue-us-ny',
+  'venue-ua',
+  'venue-za'
```
dianashk: woah! that's a whole lotta bundles!!! hooray for 👜👜👜 of data

PR #138, opened 2016-09-23
switched from `edtf:superseded` to `wof:superseded_by`

trescube:
The `wof:superseded_by` field is more reliable than the `edtf:superseded` field. There are 789 neighbourhood records with a non-empty `wof:superseded_by` field but no `edtf:superseded` field.

Connected to pelias/pelias#413


dianashk:
LGTM

PR #139, opened 2016-09-26
use `wof:label` for name instead of `wof:name` when available

### LINKED ISSUE #134: Use `wof:label` if available for record name
The WOF team has been updating records to include a `wof:label` field that should be used for name.  While importing, use the `wof:label` if available, falling back to `wof:name`.

trescube:
Fixed #134

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -46,6 +46,14 @@ function getBoundingBox(properties) {
   }
 }
 
+function getName(properties) {
+  if (properties.hasOwnProperty('wof:label')) {
+    return properties['wof:label'];
+  } else {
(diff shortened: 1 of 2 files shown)


trescube:
I found an example where this isn't ideal:

https://whosonfirst.mapzen.com/spelunker/id/101723183/#10/36.1869/-86.7852

This is the locality record for Nashville, TN (a pretty big city).  The `wof:name` is Nashville but the `wof:label` is Nashville-Davidson County.  See https://github.com/whosonfirst-data/whosonfirst-data/issues/482

trescube:
Don't merge until we get an answer on the Nashville issue.

PR #142, opened 2016-10-03
Read venue import and download settings from config

orangejulius:
Both the downloader script and importer now read the [venues config setting](https://github.com/pelias/config/pull/35) to decide whether or not to import venues. The corresponding PR to pelias/config makes the venue import default to false, so merging these two PRs will also disable importing venues unless configured otherwise.


trescube:
:shipit:

dianashk:
LGTM

PR #157, opened 2016-10-17
Remove parent id hierarchy support

### LINKED ISSUE #156: Remove support for hierachy-by-parent-id
Since we've switched to establishing parentage by navigating the array of hierarchies, storing `parent_id` and creating hierarchy by chaining `parent_id` values are both deprecated and can be removed.

trescube:
`parent_id` hierarchy walking is grossly outdated.  Fixes #156


orangejulius:
Woah, we were walking the whole hierarchy when there was one already calculated for us? I assume this was done because the hiearchies weren't great when we started using WOF (I also assume they're much better now). This definitely feels better, and is probably faster too! :shipit:

trescube:
When we first wrote the importer, we just went with walking up the hierarchy implied by `wof:parent_id` values.  We finally worked out the kinks of using the `wof:hierarchies` so the `wof:parent_id` code was left in place though it wasn't being used.

Issue #146, opened 2016-10-11
Data download script should create meta directory


orangejulius:
If the wof download directory as configured in `pelias.json` does not contain a  `meta` directory, the WOF downloader script will fail with an unhelpful error message. It's easy enough to create a directory in node.js, so the downloader should probably do this.

PR #158, opened 2016-10-20
Create meta directory before running downloader

orangejulius:
This directory needs to exist to put the meta files into.

Fixes https://github.com/pelias/whosonfirst/issues/146

### DIFF
--- a/download_data.js
+++ b/download_data.js
@@ -1,9 +1,13 @@
 var child_process = require('child_process');
 var async = require('async');
+var fs = require('fs-extra');
 var os = require('os');
 
 var bundles = require('./src/bundleList');
 var config = require('pelias-config').generate();
+
+//ensure required directory structure exists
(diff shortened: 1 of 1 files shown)


download_data.js
```diff
@@ -1,10 +1,14 @@
 var child_process = require('child_process');
 var async = require('async');
+var fs = require('fs');
 var os = require('os');
 
 var bundles = require('./src/bundleList');
 var config = require('pelias-config').generate();
 
+//ensure required directory structure exists
+fs.mkdirSync(config.imports.whosonfirst.datapath + '/meta');
```
trescube: `mkdirSync` errors if the directory already exists, use [`ensureDIrSync`](https://www.npmjs.com/package/fs-extra#ensuredirdir-callback) instead

orangejulius: ooh good catch!

PR #160, opened 2016-10-20
Remove iso2 support

trescube:
`country` records now use `wof:country` for ISO which is then used when traversing the hierarchies for setting `alpha3`


trescube:
The only place wof-pip-service uses `iso:country` is to use the appropriate US county name.

PR #163, opened 2016-10-24
Use Async methods for file loading

orangejulius:
A while back I noticed that the whosonfirst importer was not able to reach 100% CPU usage while running. It turns out because it used synchronous file loading methods it was spending a lot of time waiting. This PR uses async file loading with a limit of 10 concurrent disk operations to remove that bottleneck. It also removes the `fileIsReadable` stream which was also reading files from disk a second time, because that doesn't seem very efficient either. @trescube do you remember why that's in there? Is there a specific case it handles that reading files in cannot?

Via benchmarking it looks like this branch is about [25% faster](https://docs.google.com/spreadsheets/d/11mh5P56Q7xULkM7kiEaUsCqX9O8tG1Em-xjc_0ibzJ0/edit#gid=0)! The importer currently runs in around 8 minutes, which is pretty fast, but once we add 15 million venues, 25% is a nice saving!


trescube:
I don't remember why `fileIsReadable` was needed.  Probably chalk it up to my new-ness to node.

trescube:
On second thought, I think `fileIsReadable` had something to do with git-lfs issues.

orangejulius:
It looks like the `git-lfs` issues are handled (or at least logged) in the loadJSON function. The `fileIsReadable` stream seemed to basically just filter out nonexistent files, which is handled by the `if (err) { next } ...` part of the callback now

PR #182, opened 2016-12-21
added schema validation and updated lint configuration

trescube:
* use Joi for sub-config schema validation, incorporated into download_data.js and import.js
* copied .jslintrc from latest api

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,10 +1,22 @@
 {
   "node": true,
-
   "curly": true,
-  "latedef": true,
-  "quotmark": true,
+  "eqeqeq": true,
+  "esversion": 6,
+  "freeze": true,
+  "immed": true,
+  "indent": 2,
+  "latedef": false,
+  "newcap": true,
+  "noarg": true,
+  "noempty": true,
+  "nonbsp": true,
+  "nonew": true,
+  "plusplus": false,
+  "quotmark": "single",
   "undef": true,
-  "unused": true,
-  "trailing": true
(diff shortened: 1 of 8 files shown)


src/components/loadJSON.js
```diff
@@ -5,19 +5,21 @@ var maxInFlight = 10;
...
           next(null, object);
         } catch (parse_err) {
-          console.error('exception on %s:', record.path, parse_err);
+          console.error('exception parsing JSON in file %s:', record.path, parse_err);
           console.error('Inability to parse JSON usually means that WOF has been cloned ' +
                         'without using git-lfs, please see instructions here: ' +
                         'https://github.com/whosonfirst/whosonfirst-data#git-and-large-files');
-          next(null, {});
+          next(parse_err);
         }
```
orangejulius: this all changes the behavior here. It's actually a change we want to make, but maybe it should be in a separate PR?

trescube: So make lint updates in a separate PR?

orangejulius: No, lint updates are fine to change. But the changes to this file change the behavior, because any error will now stop the stream. We want to do that (As mentioned in [this comment](https://github.com/pelias/pelias/issues/477#issuecomment-265002768)), but it seems better to make that change in its own PR.

trescube: Lint is complaining about the long lines, that's why I changed them.  I'll disable that in .jslintrc for now.

PR #192, opened 2017-01-26
inject schema to config.generate

trescube:


src/bundleList.js
```diff
@@ -3,10 +3,10 @@
 // for info on what each bundle type means
 //
 // venue bundle data has to be imported only after all hierarchy bundles are done
-// 
+//
 // downloading can be done in any order, but the same order might as well be used
 var hierarchyBundles = [
-  'continent',
+  // 'continent',
```
orangejulius: why you hate continents?

trescube: d'oh!  fixed

orangejulius: If you get a chance, use rebase to squash that second commit where you add continents back into the first one. Otherwise we'll forever lose in git blame who actually added them, and when :P

PR #196, opened 2017-02-06
Add postalcodes to the import process

dianashk:
- [x] switched to using index.txt from the bundles server to compile a list of available bundles
- [x] add postalcode to download and import processes

Connected to pelias/api#350


src/bundleList.js
```diff
@@ -17,286 +42,39 @@ var hierarchyBundles = [
...
-  'venue-us-sd',
-  'venue-us-vt',
-  'venue-us-nj',
-  'venue-us-wv',
-  'venue-us-il',
-  'venue-us-or',
-  'venue-us-wy',
-  'venue-us-ny',
-  'venue-ua',
-  'venue-za'
```
orangejulius: yesss

package.json
```diff
@@ -30,14 +30,15 @@
...
+    "download-file-sync": "^1.0.4",
     "iso3166-1": "^0.2.5",
     "joi": "^10.0.6",
     "lodash": "^4.5.1",
     "node-version-checker": "^2.0.0",
     "pelias-config": "2.7.1",
     "pelias-dbclient": "2.0.0",
     "pelias-logger": "0.1.0",
-    "pelias-model": "4.4.0",
+    "pelias-model": "git://github.com/pelias/model.git#add-postalcode",
```
orangejulius: don't forget to change this back. Ideally it would be rebased out, otherwise on the off chance we want to look at the history of the dependencies, a much more thorough investigation will be required to know when this version was last changed from an actual version, rather than flipped around for testing.

dianashk: yup, will definitely update as soon as the greenkeeper PRs roll in :)

PR #204, opened 2017-03-14
Postalcodes should not be in the default bundle list

dianashk:
Not everyone wants to run imports or downloads with postalcodes enabled. In fact, our admin-lookup code doesn't use postalcodes at this time and currently has no way to limit the download to admin-only without postalcodes.

This change introduces a new pelias config parameter, `imports.whosonfirst.importPostalcodes` that will be set to false by default.

***IMPORTANT*** Before merging this, make sure there is a new parameter in our whosonfirst import chef script that sets the config value of `imports.whosonfirst.importPostalcodes` to true.


src/bundleList.js
```diff
@@ -6,10 +6,7 @@ const path = require('path');
 const downloadFileSync = require('download-file-sync');
 const _ = require('lodash');
 
-const peliasConfig = require('pelias-config').generate();
-
-// validate the WOF importer configuration before continuing
-require('./configValidation').validate(peliasConfig.imports.whosonfirst);
+const peliasConfig = require( 'pelias-config' ).generate(require('../schema'));
```
trescube: I don't think you need to validate against the schema here since this is only used in download_data.js and import.js which already validate.

trescube:
Nice tests and thanks for switching passing the schema to `generate()`!

Issue #209, opened 2017-04-04
cannot import data


cellum:
Downloading is not problem, "node download_data.js" will download all data, but import gives error:
Error: ENOENT: no such file or directory, open '/opt/data/110/869/286/9/1108692869.geojson'
    at Error (native)

..and there is more missing .geojson files..

oliverbienert:
I have the same issue. Download with `npm run download -- --admin-only` prints a long list of bundle file names to console and ends with `All done downloading WOF!`  
Unfortunately, running `npm start` gives the error mentioned by @cellum.
And, yes, the file is not there, directory is empty.

trescube:
2 things:

- I've opened an [issue](https://github.com/whosonfirst-data/whosonfirst-data/issues/719) with the Who's on First team
- I'm working on fixes to the whosonfirst importer and wof-admin-lookup to not fail catastrophically if a bundle listed in the bundle isn't actually there.

dianashk:
@oliverbienert, @cellum: have you been able to get past this? Is there anything else we can do to help? If this has been resolved, I'd like to close the issue.

Thanks! :wave:

orangejulius:
It looks like this was a transient issue with WOF data. Let us know if it's still happening.

PR #215, opened 2017-04-12
Update pelias-parallel-stream to 0.1.0

orangejulius:
This was missed by greenkeeper, which appears to be happening a lot lately. Maybe it's something we should look into.


trescube:
Sweet!  Now I can make changes in https://github.com/pelias/whosonfirst/issues/209#issuecomment-291970842

PR #229, opened 2017-05-03
Make missing data file non-fatal (among other things)

trescube:
This ended up being a refactoring but the major drive was to make missing data file non-fatal when configured that way.

`missingFilesAreFatal` has been added to schema and is passed to loadJSON.  If `true`, missing data files kill the entire import.  If `false`, a missing data files are logged but import continues.  

It ended up requiring pelias/config#57 but the code is much cleaner now and there are more tests.  As a side effect, bundleList.js now operates correctly since `importVenues` and `importPostalcodes` could have `yes`/`no` values which always evaluated to `true` since the types weren't being massaged.  

`isValid` ended up being renamed to `isNotNullIsland` since the entire purpose of the filter was to exclude Null Island (the country) and not all records with ids having fewer than 6 digits.  In reality, there's no change since no WOF records have ids with fewer than 6 digits but this change is more accurate.  

The entire pelias-config `imports.whosonfirst` blob is now injecting into readStream instead of piece meal values.  

readStream tests now use [temp](https://www.npmjs.com/package/temp) to handle temp directory operations.


src/components/isNotNullIsland.js
```diff
@@ -0,0 +1,8 @@
+const filter = require('through2-filter');
```
dianashk: It might be better to name this stream something like is NotBlacklisted which would be more generic and allow us to add more IDs to the list in the future. I presume null island itself would get taken care of by the check you added in the other PR about not visiting null island. If that's true, this stream might not be necessary at this time and would save a little more time during import.

PR #231, opened 2017-05-04
Move null island postalcodes filter

### LINKED ISSUE #228: Exclude null island postal codes earlier in import process
There's a filter that excludes placetype=postalcode records with lat/lon=0/0 but it happens *after* loadJSON (the most expensive part) and a few other steps when that filter could happen much earlier since the determining factors are available from the meta files.

trescube:
Fixes #228 

Moved `recordNotVisitingNullIsland` filter stream to run before the expensive `loadJSON`, shaving about 1 minute off of the import time for US/GB postalcodes (14 minutes -> 13 minutes).


src/components/recordNotVisitingNullIsland.js
```diff
@@ -1,16 +1,18 @@
-var filter = require('through2-filter');
-var _ = require('lodash');
+const filter = require('through2-filter');
```
dianashk: I don't see where this stream is being used. Did it sneak in during the commit or am I missing something?

trescube: Hmm... I could've sworn I deleted that.

PR #233, opened 2017-05-15
exposed `metadataStream` so users don't have to know how WOF root is …

trescube:
…laid out


index.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
+  metadataStream: require('./src/components/metadataStream'),
+  extractFields: require('./src/components/extractFields').create,
```
dianashk: adding extractFields but not giving it the same testing treatment as metadata. Any reason for that?

PR #243, opened 2017-06-28
Filtered download

dianashk:
Connected to https://github.com/pelias/pelias/issues/506

### Overview
Adds the ability to download only data related to a specified WOF record.

### Things to note
- new configuration params
      - `importers.whosonfirst.importPlace`: id of place to filter the data by, optional
      - `importers.whosonfirst.api_key`: whosonfirst.mapzen.com api key to be used for descendant and place info queries, optional
- download script gets called the same as before and is entirely config driven


orangejulius:
Feedback so far:

First of all this is really cool

* It needs a limit on the number of download processes it will run. It was spawning over 200 processes which my laptop did not appreciate  :) Then again I was downloading all data for the USA which is quite a lot. Should we add a note to the readme that it's intended to be used for city sized areas?
* Does Joi allow importPlace be either a string _or_ an integer? I initially got an error since I put it as an integer

src/components/loadJSON.js
```diff
@@ -8,6 +8,11 @@ const logger = require( 'pelias-logger' ).get( 'whosonfirst' );
 
 module.exports.create = function create(wofRoot, missingFilesAreFatal) {
   return parallelStream(maxInFlight, function(record, enc, next) {
+
+    if (!record.path || record.path === 'path') {
```
trescube: Could a warning message be added here so we know that something went wrong?

README.md
```diff
@@ -50,8 +50,10 @@ The following configuration options are supported by this importer.
 | key | required | default | description |
 | --- | --- | --- | --- |
 | `imports.whosonfirst.datapath` | yes | | full path to where Who's on First data is located (note: the included [downloader script](#downloading-the-data) will automatically place the WOF data here, and is the recommended way to obtain WOF data) |
+| `imports.whosonfirst.api_key` | no | | used by the filtered download script, must be set if using `imports.whosonfirst.importPlace` config option |
 | `imports.whosonfirst.importPostalcodes` | no | false | set to `true` to include postalcodes in the data download and import process |
 | `imports.whosonfirst.importVenues` | no | false | set to `true` to include venues in the data download and import process |
+| `imports.whosonfirst.importPlace` | no | | set to a WOF id string indicating the region of interest, only data pertaining to that place shall be downloaded  |
```
tigerlily-he: Link to the WOF Spelunker to obtain the WOF id: https://whosonfirst.mapzen.com/spelunker/

Is it possible to specify an array (ex. two countries)?

missinglink: would also be :+1: to have more details in the readme about how to get your own whosonfirst `api_key`, for new users this wouldn't be intuitive.

missinglink:
I ran this code today and it worked great, one thing which would be more convenient is what Lily mentioned, for the openstreetmap `end-to-end` tests we have combined `vancouver` and `queens`.

I tried changing the ID in `~/pelias.json` and re-running the script, however that seemed to delete the files downloaded the first time, would it be possible to have the script be non-destructive (leaving existing files in place)?

tigerlily-he:
If someone downloads whosonfirst data for New York City only and then imports OpenAddresses data for Philadelphia, it would be nice to have a warning that admin-lookup won't work because there isn't any WOF data for Philadelphia.

dianashk:
### Updates
- more info, links, and warnings in README, as requested by everyone
- fix the filtered download script to merge with existing metafiles so users can run the script multiple  times with different ids, as @missinglink tried to do

### Notes
- I'd like to leave the `importPlaces` config option a single id and we can put in an issue to allow for arrays in the future

utils/download_data_filtered/downloadPlaces.js
```diff
@@ -0,0 +1,97 @@
...
+const child_process = require('child_process');
+const path = require('path');
+const fs = require('fs-extra');
+const async = require('async');
+const logger = require('pelias-logger').get('download_data_filtered');
+const parallelStream = require('pelias-parallel-stream');
+const streamArray = require('stream-array');
+const os = require('os');
+
+const maxInFlight = os.cpus().length * 10;
```
orangejulius: This value is really high. Many/most developer machines these days are quad core with hyperthreading, so `os.cpus().length` will return 8, so there would be up to 80 downloads happening at once. Maybe up to 4 concurrent downloads can help ensure that network bandwidth is used effectively, but 80 is just overkill. The number of CPUs isn't really important either, so I'd suggest hardcoding this to between 2 and 4. Things should happen just as fast if not faster, and with much less thrashing of processes on users machine (mine is almost unusable when a download is happening).

src/components/loadJSON.js
dianashk: added below the line

PR #246, opened 2017-07-08
fix(package): update pelias-model to version 4.10.1

trescube:
Closes #245


trescube:
I'm going to merge this because it was greenkeepers' anyway.

Issue #249, opened 2017-07-10
Encoding error during import


vincbon:
An encoding error occurs right when I start the import. I have tried downloading the data again (with the script), same thing. Maybe it's the connection : data is downloaded at about 300kb/s and the final size is 8Gb (hierarchy only) as opposed to the "tens of GB" descriped here.
I'm on Mac too.

Here's the output :

pelias-whosonfirst@0.0.0-semantic-release start /Users/vincent.bonhomme/Documents/pelias/whosonfirst
node import.js

2017-07-10T08:00:49.140Z - info: [whosonfirst] Loading wof-continent-latest.csv records from /Users/vincent.bonhomme/Downloads/data/whosonfirst/meta
buffer.js:557
  throw new TypeError('Unknown encoding: ' + encoding);
  ^

TypeError: Unknown encoding: 
    at stringSlice (buffer.js:557:9)
    at Buffer.toString (buffer.js:593:10)
    at CSVStream.write (/Users/vincent.bonhomme/Documents/pelias/whosonfirst/node_modules/csv-stream/index.js:59:34)
    at ReadStream.ondata (_stream_readable.js:628:20)
    at emitOne (events.js:115:13)
    at ReadStream.emit (events.js:210:7)
    at addChunk (_stream_readable.js:252:12)
    at readableAddChunk (_stream_readable.js:239:11)
    at ReadStream.Readable.push (_stream_readable.js:197:10)
    at onread (fs.js:2004:12)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! pelias-whosonfirst@0.0.0-semantic-release start: `node import.js`
npm ERR! Exit status 1

missinglink:
hi @vincbon, that's unusual.

it's possible that the data has been corrupted during download, but you said you tried twice, so that's fairly unlikely.

I'm assuming you're not getting any errors printed to the console and your HDD is not full :)

I have a copy I downloaded using that script last week:

```bash
$ du -sh whosonfirst-data
9.5G	whosonfirst-data
```

```bash
$ head whosonfirst-data/meta/wof-continent-latest.csv 
bbox,cessation,country_id,deprecated,file_hash,fullname,geom_hash,geom_latitude,geom_longitude,id,inception,iso,iso_country,lastmodified,lbl_latitude,lbl_longitude,locality_id,name,parent_id,path,placetype,region_id,source,superseded_by,supersedes,wof_country
[...]

PR #251, opened 2017-07-11
Clarify Node.js version requirements

orangejulius:
Hopefully we can support newer versions soon (https://github.com/pelias/pelias/issues/612)

Connects https://github.com/pelias/whosonfirst/issues/249


dianashk:
Could we make it even more official by specifying engine versions in the package file?

orangejulius:
Good idea. Done

PR #264, opened 2017-08-14
ingore timestamped bundles

missinglink:
issue reported in https://github.com/pelias/whosonfirst/issues/263

```javascript
filename = 'wof-locality-20170810T000409-bundle.tar.bz2';

filename.replace(/-\d{8}T\d{6}-/, '-latest-').replace('-bundle.tar.bz2', '.csv');
"wof-locality-latest.csv"
```

note: I also renamed the variable `targetPath` to `csvFilename` for better clarity of it's intended purpose.

resolves https://github.com/pelias/whosonfirst/issues/263

PR #274, opened 2017-08-24
Streamline bundle list creation in light of timestamped bundles

dianashk:
This cleans up the way we check the bundle list for existing meta files to support the new timestamped bundles as well as the filtered download mode where no timestamps are present.


dianashk:
Connected to https://github.com/pelias/whosonfirst/pull/264

PR #280, opened 2017-09-13
Add more placetypes to download

trescube:
Add `empire`, `ocean`, and `marinearea` to download.  Also synthesizes a hierarchy from the id and placetype when no hierarchies are in the source data.  This is useful for top-level placetypes like oceans that have no hierarchies.  Without this, the ES document would not have `ocean`, `ocean_id`, and `ocean_a` with is inconsistent with lower level placetypes like neighbourhoods where the hierarchy ensures that `neighbourhhood`, `neighbourhood_id`, and `neighbourhood_a` are in the ES documents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,15 @@ Currently, the supported hierarchy types are:
 - county
 - dependency
 - disputed
+- [empire](https://www.youtube.com/watch?v=-bzWSJG93P8)
 - localadmin
 - locality
 - macrocounty
 - macrohood
 - macroregion
+- marinearea
 - neighbourhood
+- ocean
 - region
 - postalcodes (optional, see configuration)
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -31,12 +31,15 @@ Currently, the supported hierarchy types are:
 - county
 - dependency
 - disputed
+- [empire](https://www.youtube.com/watch?v=-bzWSJG93P8)
```
orangejulius: lol

package.json
```diff
@@ -41,7 +41,7 @@
     "pelias-config": "2.12.0",
     "pelias-dbclient": "2.2.2",
     "pelias-logger": "0.2.0",
-    "pelias-model": "5.0.1",
+    "pelias-model": "pelias/model",
```
orangejulius: whats this?

orangejulius: Oh you're probably waiting for this to be merged, right? https://github.com/pelias/model/pull/75

trescube: Correct, I'm stalling on merging that for a bit as I worked out how this effects the whole ecosystem.

PR #324, opened 2018-02-27
extract regional data from wof sqlite db

missinglink:
READY FOR REVIEW

this is a first pass at extracting data from the [wof sqlite databases](https://dist.whosonfirst.org/sqlite/) as an alternative to using the currently out-of-action WOF API.

I have tested this and it works great, we still need to add a little polish to it.

No additional properties are required in `pelias.json`, additionally `api_key` is no longer required.

The script also supports `postalcode`, `venue`, `constituency` and `intersection` placetypes where available.

npm `better-sqlite3` requires node 6+, is this an issue?
note: running this on node4 or less crashes node with the message `FATAL ERROR: v8::ToLocalChecked Empty MaybeLocal.`

```
"whosonfirst": {
  "datapath": "/tmp/wof",
  "importVenues": true,
  "importPostalcodes": true,
  "importConstituencies": true,
  "importIntersections": true,
  "importPlace": "101748799"
},
```

as usual, you can run the download with `npm run download`

### DIFF
--- a/schema.js
+++ b/schema.js
@@ -7,10 +7,11 @@ const Joi = require('joi');
 // optional:
 // * imports.whosonfirst.importVenues (boolean) (default: false)
 // * imports.whosonfirst.importPostalcodes (boolean) (default: false)
+// * imports.whosonfirst.importConstituencies (boolean) (default: false)
+// * imports.whosonfirst.importIntersections (boolean) (default: false)
 // * imports.whosonfirst.importPlace (string) (default: none)
-// * imports.whosonfirst.api_key (string) (default: none)
+// * imports.whosonfirst.api_key (string) (default: none) DEPRECATED
 // * imports.whosonfirst.missingFilesAreFatal (boolean) (default: false)
-// * imports.whosonfirst.sqliteDatabases array of objects containing filename (optional)
 
 module.exports = Joi.object().keys({
   imports: Joi.object().keys({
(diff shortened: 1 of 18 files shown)


orangejulius:
Code looks simple and clear! SQLite definitely for the win here. One nice thing to add would be to update the downloader to use the SQLite files and run the extract when a filtering id is specified.

utils/sqlite_extract_data.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
```
orangejulius: `use strict` is only needed on Node.js 4, which we're now dropping support for with the addition of better-sqlite3, right?

schema.js
```diff
@@ -19,7 +19,10 @@ module.exports = Joi.object().keys({
       api_key: Joi.string(),
       importVenues: Joi.boolean().default(false).truthy('yes').falsy('no').insensitive(true),
       importPostalcodes: Joi.boolean().default(false).truthy('yes').falsy('no').insensitive(true),
-      missingFilesAreFatal: Joi.boolean().default(false).truthy('yes').falsy('no').insensitive(true)
+      missingFilesAreFatal: Joi.boolean().default(false).truthy('yes').falsy('no').insensitive(true),
+      sqlite_files: Joi.array().items(Joi.object().keys({
+        filename: Joi.string()
+      }).requiredKeys('filename').unknown(true)),
```
orangejulius: What would it take to remove the need for the new `sqlite_files` `pelias.json` property entirely? Previously, the importer was able to determine what combination of admin, postalcode, and venue bundles to download. Can we do the same with the SQLite downloads?

orangejulius:
I've cherry picked the first commit of this PR into https://github.com/pelias/whosonfirst/pull/325 and also merged https://github.com/pelias/whosonfirst/pull/326 which removes support for Node.js 4. A rebase should help with the test failures!

schema.js
missinglink: Good idea, that's pretty easy to implement and would be super convenient.

We'd need to talk about how the paths would work, currently `datapath` refers to the directory which has `data` and `meta` as it's children.

I have `sqlite_files.filename` set up to point to an absolute path. In order to get the downloads working, we would probably need to change that to only list the filename, not the path.

Then the question is, which path to we download to and load from? we could use `datapath` or we could introduce a new path param?

I'm currently just downloading my databases in to the `datapath` and it looks like this:

```
$ ls -1
data
meta
whosonfirst-data-latest.db
whosonfirst-data-postalcode-us-latest.db
```

It's probably fine to use `datapath` for most use-cases, we just need to make users aware that the files can be quite large.

```
15G whosonfirst-data-latest.db
1.4G whosonfirst-data-postalcode-us-latest.db
```

orangejulius: Yeah, putting the DBs into `datapath` is perfect. Eventually, if/when we switch to importing straight from sqlite, rather than generating the files from the sqlite DB, everything would already be in the right place and the `data` and `meta` directories would disappear. As for file size,  yes, the sqlite dbs are quite large, but we've already been warning people about WOF data size forever. It should be ok.

PR #338, opened 2018-03-27
Urls update

gibranparvez:
Equivalent to [this commit](https://github.com/pelias/whosonfirst/commit/b0f7de07a8082a2bf10836a4be75d6496fbc11a7#diff-ea91ea9091564ca074d00caeac418e76) from official repo.

PR #339, opened 2018-03-27
Update URLs to new WOF source

missinglink:
Commit taken from https://github.com/pelias/whosonfirst/pull/338


orangejulius:
Weird, not sure how I missed those in https://github.com/pelias/whosonfirst/pull/325. But good to fix

missinglink:
you only missed one, the other one I favoured the commit without the trailing slash :)

PR #368, opened 2018-06-22
docker: ignore node_modules

missinglink:
docker: ignore node_modules

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 .git
+node_modules


orangejulius:
Oh yeah. This is important ,to prevent whatever unique combination of local/out-of-date/modified node modules might be installed locally when building, right?

It probably doesn't matter as much since all our docker images our built on a clean slate in CircleCI, but should we go through and make sure this is in all the dockerignore files?

PR #374, opened 2018-08-05
feat(abbreviations): Use ISO-3166-1 alpha3 code from WOF records

orangejulius:
Pelias requires an ISO-3166-1 alpha3 country code on all country records. Currently, we use the `wof:country` field, which contains an alpha2 code, and convert to alpha3 codes. This PR replaces that approach and uses the `wof:country_alpha3` field, if present, to set the corresponding field on Pelias country records.

This allows us to remove the troublesome (see https://github.com/pelias/api/issues/1179) iso3166-1 NPM package.

One implication of the change is that it does mean we are trusting the data in WOF to be correct.

My read of the discussions in https://github.com/whosonfirst-data/whosonfirst-data/pull/823, https://github.com/whosonfirst-data/whosonfirst-data/pull/1055, and other related WOF issues is that this is the right field going forward. However, we can easily change it.

@nvkelso @stepps00, does this seem like the right field to use for country abbreviations?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "csv-stream": "^0.2.0",
     "download-file-sync": "^1.0.4",
     "fs-extra": "^7.0.0",
-    "iso3166-1": "^0.3.0",
     "joi": "^13.1.2",
     "klaw-sync": "^4.0.0",
     "lodash": "^4.5.1",
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -66,8 +66,8 @@ function getName(properties) {
 }
 
 function getAbbreviation(properties) {
-  if (properties['wof:placetype'] === 'country' && properties['wof:country']) {
-    return properties['wof:country'];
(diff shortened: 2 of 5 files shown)


stepps00:
That sounds like the correct approach. The ISO-3166-1 codes are stored in the `wof:country_alpha3` property. The property should be set for all `country` records, but not `dependency` records (so some lookup may still be required).

orangejulius:
After some review and further understanding from the other abbreviation PRs, it turns out this change isn't really needed.

The iso3166-1 package seems to be doing ok now (it actually pushed an update fixing some of our issues). Furthermore we would still have to convert the two letter dependency codes, so it's better to be consistent.

@stepps00 if you ever add a field with 3 letter codes for dependencies (like `PRI` instead of `PR` for [Puerto Rico](https://spelunker.whosonfirst.org/id/85633729/), for example), please let us know :)

PR #273, opened 2017-08-24
[DO NOT MERGE] switch to wof:abbreviation for dependencies

dianashk:
According to whosonfirst-data/whosonfirst-data#823, dependency placetype records are moving abbreviations from wof:country to wof:abbreviation. Don't merge until the bundles have been updated.

Previously merged prematurely via #240 and reverted via #272 

Connected to pelias/placeholder#33

PR #375, opened 2018-08-05
feat(abbreviations): use wof:abbreviation for dependencies if available

orangejulius:
This is a backwards-compatible version of https://github.com/pelias/whosonfirst/pull/273 which allows us to handle the upcoming switch from storing dependency abbreviations in the `wof:country` field to `wof:abbreviation` (see https://github.com/whosonfirst-data/whosonfirst-data/pull/823)

It should be safe to merge now and will do the right thing once WOF data is updated.

Supersedes https://github.com/pelias/whosonfirst/pull/273

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -65,14 +65,16 @@ function getName(properties) {
   }
 }
 
-function isDependencyOrCountry(placetype) {
-  return placetype === 'dependency' || placetype === 'country';
-}
-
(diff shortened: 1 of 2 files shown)


orangejulius:
Note: lets merge https://github.com/pelias/whosonfirst/pull/374 first since they touch similar parts of the code.

orangejulius:
replaced with https://github.com/pelias/whosonfirst/pull/399

PR #31, opened 2016-02-01
removed unused placetypes

trescube:
Fixed #30

PR #377, opened 2018-08-07
Add microhood

orangejulius:
This was removed in https://github.com/pelias/whosonfirst/pull/31

### DIFF
--- a/src/bundleList.js
+++ b/src/bundleList.js
@@ -31,7 +31,8 @@ const hierarchyRoles = [
   'locality',
   'borough',
   'macrohood',
-  'neighbourhood'
+  'neighbourhood',
+  'microhood'
 ];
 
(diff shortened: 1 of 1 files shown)


missinglink:
Does anyone remember why it was originally removed or know what data we are missing by not including it?

orangejulius:
I don't. This PR come from a report from gitter that some neighbourhoods in Chicago that are fairly important are stored as `microhood`s. Since they were removed 2 years ago, it's reasonble to think that more microhoods exist now than back then.

I'm still not 100% on merging this, but it might be worth investigating.

orangejulius:
The originally reported issue involved `microhood`s in admin lookup as well. I don't think we want to add support for that currently. We have already added many admin fields and things are starting to get a little cluttered.

PR #379, opened 2018-08-07
feat(download): remove obsolete download code

orangejulius:
This is the code for downloading WOF data from the Mapzen Whosonfirst
API, which has been gone since January 2018. If a similar API ever
becomes part of the WOF infrastructure, we could consider bringing back
this code.

Until then, it's best removed to avoid the suggestion that it could
still work.


orangejulius:
Oh, good catch @missinglink, we were able to remove a few pacakges!

PR #380, opened 2018-08-09
use absolute path in child process cwd

noizde:
Should resolve https://github.com/pelias/pelias/issues/736.

### DIFF
--- a/utils/sqlite_download.js
+++ b/utils/sqlite_download.js
@@ -19,7 +19,7 @@ function download(options, callback){
     return (done) => {
 
       // build shell command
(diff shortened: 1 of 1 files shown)


missinglink:
hmm.. not sure what I was thinking there....

looking at the docs for `__dirname` is says `This is the same as the path.dirname() of the __filename.` so maybe I was confused by that...

missinglink:
note: the `npm` module is not published until the code reaches the `production` branch, we will look at merging changes up to that branch very soon.

noizde:
Good to know, thanks for the heads up and the quick turnaround!

PR #389, opened 2018-09-04
Removed -bundle from wof meta filename translation.

richardcalahan:


orangejulius:
Thanks @richardcalahan. When you encountered the error what type of WOF import were you running? Was it a full WOF import or did you specify some WOF IDs to filter on via the `importPlace` `pelias.json` config variable?

richardcalahan:
It was a full import, no filters.

orangejulius:
Finally had a chance to test this on a full-WOF import, and it works great. Thanks

PR #399, opened 2018-09-19
feat(abbreviations): handle changing abbreviation fields

orangejulius:
Several abbreviation fields have changed or will change over time in WOF.

1.) `wof:shortcode` just replaced `wof:abbreviation`
2.) for dependency records, `wof:shortcode` will replace `wof:country`

This change allows us to handle both of those changing fields in a backwards compatible way, by supporting the older field but prioritizing the newer ones.

We should later come in and remove the logic to refer to the older values once we're sure they're no longer used and enough time has passed that most people will have updated their copy of WOF data.

Supersedes https://github.com/pelias/whosonfirst/pull/273
Supersedes https://github.com/pelias/whosonfirst/pull/375
Fixes https://github.com/pelias/whosonfirst/issues/398
Fixes https://github.com/pelias/pelias/issues/645
Connects https://github.com/whosonfirst/whosonfirst-properties/issues/78

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -65,15 +65,17 @@ function getName(properties) {
   }
 }
 
-function isDependencyOrCountry(placetype) {
-  return placetype === 'dependency' || placetype === 'country';
-}
-
 function getAbbreviation(properties) {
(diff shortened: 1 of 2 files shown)


missinglink:
Tested and works for me

PR #417, opened 2019-01-13
Supports for WOF SQLite database

### LINKED ISSUE #416: Using WOF SQLite database ?
It seems that the WOF team is updating SQLite databases more often than bundles https://github.com/whosonfirst/go-whosonfirst-dist/issues/9.

That's why I thought it would be interesting to supports SQLite for import and [wof-admin-lookup](https://github.com/pelias/wof-admin-lookup).

This will prevent errors like
- corrupted bundle (experienced with https://github.com/whosonfirst/go-whosonfirst-dist/issues/9)
- inodes errors (too many files when we wants to build our own bundles)
- database portability (usefull for HA)

That's an idea, if you think it's interesting, I'd be happy to implement it.

Joxit:
This PR add supports to SQLite databases.
[pelias/whosonfirst](https://github.com/pelias/whosonfirst) will still supports bundle, for sqlite import, a new options will be added : `imports.whosonfirst.sqlite` (false by default for backward compatibility).

TODO list:
- [x] SQLite read stream
- [x] List available databases
- [x] Add `imports.whosonfirst.sqlite` configuration
- [x] Add SQLite stream in import pipeline
- [x] Download SQLite database for world `utils/download_data_all.js`
- [x] Check performances between SQLite and bundle
- [x] Expose SQLite Read Stream (for [pelias/wof-admin-lookup](https://github.com/pelias/wof-admin-lookup))
- [x] Cleanup SQLite (this can take some times)
- [x] Supports importPlace

fixes #416

### DIFF
--- a/src/bundleList.js
+++ b/src/bundleList.js
@@ -41,7 +41,7 @@ const venueRoles = [
   'venue'
 ];
 
-const SQLITE_REGEX = /whosonfirst-data-[a-z-]+.db/;
+const SQLITE_REGEX = /whosonfirst-data-[a-z0-9-]+.db/;
 
 function getPlacetypes() {
   let roles = hierarchyRoles;
--- a/src/bundleList.js
+++ b/src/bundleList.js
@@ -41,7 +41,7 @@ const venueRoles = [
   'venue'
 ];
 
-const SQLITE_REGEX = /whosonfirst-data-[a-z0-9-]+.db/;
+const SQLITE_REGEX = /whosonfirst-data-[a-z0-9-]+\.db$/;
 
 function getPlacetypes() {
   let roles = hierarchyRoles;
--- a/src/bundleList.js
+++ b/src/bundleList.js
@@ -41,6 +41,7 @@ const venueRoles = [
   'venue'
 ];
 
+const SQLITE_REGEX = /whosonfirst-data-[a-z-]+.db/;
 
 function getPlacetypes() {
   let roles = hierarchyRoles;
@@ -121,6 +122,25 @@ function getBundleList(callback) {
   });
 }
 
+function getDBList(callback) {
+  const databasesPath = path.join(peliasConfig.imports.whosonfirst.datapath, 'sqlite');
+  //ensure required directory structure exists
+  fs.ensureDirSync(databasesPath);
+  const dbList = fs.readdirSync(databasesPath).filter(d => SQLITE_REGEX.test(d));
+
+  if (_.isEmpty(dbList)) {
+    return callback(`No database found in ${databasesPath}`);
+  }
+  callback(null, dbList);
+}
+
+function getList(callback) {
+  if (peliasConfig.imports.whosonfirst.sqlite) {
+    return getDBList(callback);
+  }
+  getBundleList(callback);
+}
+
 function initBundleBuckets(roles) {
   const bundleBuckets = {};
   roles.forEach( (role) => {
@@ -151,4 +171,4 @@ function combineBundleBuckets(roles, bundleBuckets) {
 }
 
 module.exports.getPlacetypes = getPlacetypes;
-module.exports.generateBundleList = getBundleList;
+module.exports.generateBundleList = getList;
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -6,14 +6,45 @@ class SQLiteStream extends Readable {
     super({ objectMode: true, autoDestroy: true });
     this._db = new Sqlite3(dbPath, { readonly: true });
     this._iterator = this._db.prepare(sql).iterate();
+    this.on('end', () => {
+      this._db.close();
+    });
   }
 
   _read() {
     const elt = this._iterator.next();
-    if(!this.push(!elt.done && elt.value ? elt.value : null)) {
-      this._db.close();
-    }
+    this.push(!elt.done && elt.value ? elt.value : null);
   }
 }
 
+/**
+ * Filter deprecated, superseded and null island objects
+ */
+function findGeoJSON() {
+  return `
+SELECT geojson.id, geojson.body
+  FROM geojson JOIN spr
+  ON geojson.id = spr.id
+  WHERE
+    geojson.id != 1
+      AND
+    spr.is_deprecated = 0
+      AND
+    spr.is_superseded = 0
+      AND
+    (spr.latitude != 0 OR spr.longitude != 0)`;
+}
+
+/**
+ * Add filter and placetypes
+ */
+function findGeoJSONByPlacetype(placetypes) {
+  return findGeoJSON() + `
+      AND
+    spr.placetype IN ('${placetypes.join('\',\'')}')
+`;
+}
+
 module.exports = SQLiteStream;
+module.exports.findGeoJSON = findGeoJSON;
+module.exports.findGeoJSONByPlacetype = findGeoJSONByPlacetype;
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -1,0 +1,19 @@
+const Readable = require('stream').Readable;
+const Sqlite3 = require('better-sqlite3');
+
+class SQLiteStream extends Readable {
+  constructor(dbPath, sql) {
+    super({ encoding: 'utf-8', autoDestroy: true });
+    this._db = new Sqlite3(dbPath, { readonly: true });
+    this._iterator = this._db.prepare(sql).iterate();
+  }
+
+  _read() {
+    const elt = this._iterator.next();
+    if(!this.push(!elt.done && elt.value ? elt.value.body : null)) {
+      this._db.close();
+    }
+  }
+}
+
+module.exports = SQLiteStream;
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -41,7 +41,9 @@ SELECT geojson.id, geojson.body
  * Add filter and placetypes
  */
 function findGeoJSONByPlacetype(placetypes) {
-  placetypes = typeof placetypes === 'string' ? [placetypes] : placetypes;
+  if(!Array.isArray(placetypes)) {
+    placetypes = [ placetypes ];
+  }
   return findGeoJSON() + `
       AND
     spr.placetype IN ('${placetypes.join('\',\'')}')
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -3,14 +3,14 @@ const Sqlite3 = require('better-sqlite3');
 
 class SQLiteStream extends Readable {
   constructor(dbPath, sql) {
-    super({ encoding: 'utf-8', autoDestroy: true });
+    super({ objectMode: true, autoDestroy: true });
     this._db = new Sqlite3(dbPath, { readonly: true });
     this._iterator = this._db.prepare(sql).iterate();
   }
 
   _read() {
     const elt = this._iterator.next();
-    if(!this.push(!elt.done && elt.value ? elt.value.body : null)) {
+    if(!this.push(!elt.done && elt.value ? elt.value : null)) {
       this._db.close();
     }
   }
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -5,6 +5,14 @@ const logger = require('pelias-logger').get('whosonfirst:sqliteStream');
 class SQLiteStream extends Readable {
   constructor(dbPath, sql) {
     super({ objectMode: true, autoDestroy: true, highWaterMark: 32 });
+
+    // ensure indices exist
+    // note: this can be removed once the upstream PR is merged:
+    // https://github.com/whosonfirst/go-whosonfirst-sqlite-features/pull/4
+    new Sqlite3(dbPath)
+      .exec('CREATE INDEX IF NOT EXISTS spr_obsolete ON spr (is_deprecated, is_superseded)')
+      .close();
+
     this._db = new Sqlite3(dbPath, { readonly: true });
     this._iterator = this._db.prepare(sql).iterate();
     this.on('error', (e) => { logger.error(e); });
@@ -29,16 +37,14 @@ function findGeoJSON() {
 SELECT geojson.id, geojson.body
   FROM geojson JOIN spr
   ON geojson.id = spr.id
-  WHERE
-    geojson.id != 1
-      AND
-    spr.is_deprecated = 0
-      AND
-    spr.is_superseded = 0
-      AND
-    spr.name <> ''
-      AND
-    (spr.latitude != 0 OR spr.longitude != 0)`;
+  WHERE geojson.id != 1
+  AND spr.is_deprecated = 0
+  AND spr.is_superseded = 0
+  AND NOT TRIM( IFNULL(spr.name, '') ) = ''
+  AND NOT (
+    spr.latitude = 0 AND
+    spr.longitude = 0
+  )`;
 }
 
 /**
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -1,19 +1,23 @@
 const Readable = require('stream').Readable;
 const Sqlite3 = require('better-sqlite3');
+const logger = require('pelias-logger').get('whosonfirst:sqliteStream');
 
 class SQLiteStream extends Readable {
   constructor(dbPath, sql) {
-    super({ objectMode: true, autoDestroy: true });
+    super({ objectMode: true, autoDestroy: true, highWaterMark: 32 });
     this._db = new Sqlite3(dbPath, { readonly: true });
     this._iterator = this._db.prepare(sql).iterate();
-    this.on('end', () => {
-      this._db.close();
-    });
+    this.on('error', (e) => { logger.error(e); });
+    this.on('end', () => { this._db.close(); });
   }
 
   _read() {
-    const elt = this._iterator.next();
-    this.push(!elt.done && elt.value ? elt.value : null);
+    var ok = true;
+    while(ok){
+      const elt = this._iterator.next();
+      if (!elt.done) { ok = this.push(elt.value); }
+      else { this.push(null); break; }
+    }
   }
 }
 
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -32,6 +32,8 @@ SELECT geojson.id, geojson.body
       AND
     spr.is_superseded = 0
       AND
+    spr.name <> ''
+      AND
     (spr.latitude != 0 OR spr.longitude != 0)`;
 }
 
@@ -39,6 +41,7 @@ SELECT geojson.id, geojson.body
  * Add filter and placetypes
  */
 function findGeoJSONByPlacetype(placetypes) {
+  placetypes = typeof placetypes === 'string' ? [placetypes] : placetypes;
   return findGeoJSON() + `
       AND
     spr.placetype IN ('${placetypes.join('\',\'')}')
--- a/src/components/sqliteStream.js
+++ b/src/components/sqliteStream.js
@@ -46,10 +46,29 @@ function findGeoJSONByPlacetype(placetypes) {
   }
   return findGeoJSON() + `
       AND
-    spr.placetype IN ('${placetypes.join('\',\'')}')
-`;
+    spr.placetype IN ('${placetypes.join('\',\'')}')`;
+}
+function findGeoJSONByPlacetypeAndWOFId(placetypes, wofids) {
+  if(!Array.isArray(wofids)) {
+    wofids = [ wofids ];
+  }
+  wofids = wofids.map(id => parseInt(id, 10)).join(',');
+  return findGeoJSONByPlacetype(placetypes) + `
+      AND
+    spr.id IN (
+      SELECT DISTINCT id
+        FROM ancestors
+        WHERE id IN (${wofids})
+      UNION SELECT DISTINCT id
+        FROM ancestors
+        WHERE ancestor_id IN (${wofids})
+      UNION SELECT DISTINCT ancestor_id
+        FROM ancestors
+        WHERE id IN (${wofids})
+    )`;
 }
 
 module.exports = SQLiteStream;
 module.exports.findGeoJSON = findGeoJSON;
 module.exports.findGeoJSONByPlacetype = findGeoJSONByPlacetype;
+module.exports.findGeoJSONByPlacetypeAndWOFId = findGeoJSONByPlacetypeAndWOFId;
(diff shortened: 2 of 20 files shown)


src/bundleList.js
```diff
@@ -41,6 +41,7 @@ const venueRoles = [
   'venue'
 ];
 
+const SQLITE_REGEX = /whosonfirst-data-[a-z-]+.db/;
```
missinglink: Should this regex be able to match any WOF sqlite db?

```js
/whosonfirst-data-[a-z-]+.db/.test('https://dist.whosonfirst.org/sqlite/whosonfirst-data-constituency-us-ct-1481486175.db')

fals
```

src/components/sqliteStream.js
```diff
@@ -0,0 +1,19 @@
...
+
+  _read() {
+    const elt = this._iterator.next();
+    if(!this.push(!elt.done && elt.value ? elt.value : null)) {
+      this._db.close();
+    }
+  }
+}
+
+module.exports = SQLiteStream;
```
missinglink: This is really nice :)

src/bundleList.js
Joxit: Nice catch ! This one will not work :disappointed: I change the regex.

Joxit:
I have a question, should a provide a script for database optimisation/size reduce ?
Because when we download/build the database from WOF, we have tons of useless data and we can reduce the database size by removing useless entries in `geojson`/`spr`/`ancestors` and droping `names` and `concordances`.
Maybe this is interesting only for those who make their own WOF database and can differ between use cases :confused:.

Joxit:
I did some performance tests for SQLite versus meta/bundle on a world instance....
The fact is SQLite is faster than meta/bundle because bundle is limited by IO of tiny files.

When I use filtered queries (by placetypes), meta/bundle is faster than SQLites for placetypes from oceans (7 records in 7s vs 52s) to country (210 records 23s vs 67s).
But SQLite is faster from region (4k records 96s vs 115) to locality (163k records 1973s vs ?)

Here is the full result

| place type | size | time (meta/bundle) | time (SQLite) |
| ----- | ---- | ---- | ---- |
| ocean | 7 | 3s | 52s |
| continent | 8 | 3s | 54s |
| empire | 11 | 5s | 52s |
| dependency | 40 | 8s | 52s |
| macroregion | 94 | 11s | 57s |
| country | 210 | 23s | 67s |
| borough | 241 | 13s | 53s |
| marinearea | 305 | 9s | 52s |
| macrocounty | 371 | 12s | 54s |
| region | 4k | 115s | 96s |
| county | 41k | 451s | 147s |
| localadmin | 112k | 922s | 150s |
| locality | 163k | | 1973s |

The benchmark was performed on a public cloud.

Joxit:
Hi @missinglink @orangejulius,

I think this part is now OK

src/components/sqliteStream.js
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Add filter and placetypes
+ */
+function findGeoJSONByPlacetype(placetypes) {
+  placetypes = typeof placetypes === 'string' ? [placetypes] : placetypes;
+  return findGeoJSON() + `
+      AND
+    spr.placetype IN ('${placetypes.join('\',\'')}')
+`;
+}
```
missinglink: This function is a `module.export`, so it might pay to be stricter about argument checking and casting here.

With the current code, it's possible to provide a `number` or `undefined` or `null` which will then have `.join()` called on it, potentially resulting in a fatal program error and `exit(1)`.

I personally find the ternary syntax hard to read, wrapping the expression in brackets helps avoid error and make it easier for others to read it.

I would write that line as:
```if( !Array.isArray( placetypes ) ){ placetypes = [ placetypes ]; }```

The advantage of doing it like that is that you'll always end up with an `Array` regardless of the input type, meaning `.join()` won't fail and most types will be passed on verbatim.

File: src/components/toJSONStream.js [hunk 4]
```diff
@@ -0,0 +1,15 @@
...
+
+const maxInFlight = 10;
+
+module.exports.create = () => {
+  return parallelTransform(maxInFlight, function(record, next) {
+    try {
+      next(null, JSON.parse(record.body));
+    } catch (parse_err) {
+      logger.error(`exception parsing JSON for id ${record.id}: ${parse_err}`);
+      next(parse_err);
```
missinglink: This will cause the whole importer to fail at this point and the process to `exit(1)`, is this your intention?

It might be preferable to logging a warning if one record is malformed rather than aborting the whole import.

src/readStream.js
```diff
@@ -22,6 +25,15 @@ function getMetaFilePaths(wofRoot, bundles) {
   });
 }
 
+/*
+ * Convert a base directory and list of types into a list of sqlite file paths
```
missinglink: This comment says 'list of types', is that the correct wording for this?

src/readStream.js
```diff
@@ -63,9 +92,16 @@ function createReadStream(wofConfig, types, wofAdminRecords) {
   const wofRoot = wofConfig.datapath;
   const metaFilePaths = getMetaFilePaths(wofRoot, types);
 
-  return createMetaRecordStream(metaFilePaths, types)
-  .pipe(isNotNullIslandRelated.create())
-  .pipe(loadJSON.create(wofRoot, wofConfig.missingFilesAreFatal))
+  // Select correct stream between meta and SQLite based on config and do specialized stuff
+  const stream = wofConfig.sqlite ?
```
missinglink: This ternary checks that `wofConfig.sqlite` is truthy, so a value such as "false" or "no" would also trigger this behaviour.

It would be more robust if the check was `wofConfig.sqlite === true`, or checked against a specific subset of truthy values.

src/components/sqliteStream.js
```diff
@@ -0,0 +1,53 @@
...
+    geojson.id != 1
+      AND
+    spr.is_deprecated = 0
+      AND
+    spr.is_superseded = 0
+      AND
+    spr.name <> ''
+      AND
+    (spr.latitude != 0 OR spr.longitude != 0)`;
+}
```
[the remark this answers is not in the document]
missinglink: Worth noting that this uses the [json1](https://www.sqlite.org/json1.html) extension, I've found that this is included in every distribution of sqlite I've used, and has been in the `amalgamation` for some time (including very old versions).

src/readStream.js
Joxit: Okay, I changed this with `wofConfig.sqlite === true`, Joi will do the job for truthy values

src/components/toJSONStream.js
Joxit: This was not my intention, thx again :smile:

src/components/sqliteStream.js
Joxit: Updated with more test for this part :+1:

src/components/sqliteStream.js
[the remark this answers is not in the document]
Joxit: I created my own SQLite and bundle from WOF's github repositories and with their utilities. 

When I did the benchmark imports, placetype count were almost equals between SQLite and Bundle. Should I run another full test for comparison ?

The advantage with spr is that the column is typed as integer, so there are less type issue. But this is not perfect: 
```
# is_deprecated is not OK with -1 values...
sqlite> select count(*), is_deprecated from spr group by is_deprecated;
16|-1
8709067|0
13417|1
# is_superseded is OK
sqlite> select count(*), is_superseded from spr group by is_superseded;
8709235|0
13265|1
```

Also we have [isActiveRecord](https://github.com/pelias/whosonfirst/blob/0542218b98c8e505fc7c4d9b3435ff0210b2c848/src/components/isActiveRecord.js) for deprecated feature filtering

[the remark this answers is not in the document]
missinglink: I'm just a little worried that the filtering logic for the old SQLite method, the bundle method and the new SQLite method might result in different document counts?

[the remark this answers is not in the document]
Joxit: What do you mean with old SQLite method ? [this one](https://github.com/pelias/whosonfirst/blob/145e106b04dfa38954af69ed5664c375b55fba76/utils/sqlite_extract_data.js) ?
Fine, I will renew my world import and give you back the correct number of features imported.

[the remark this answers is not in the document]
Joxit:
```
sqlite> seselect count(*), json_extract(body,'$.properties."edtf:deprecated"') as edtf from geojson group by edtf;
8709067|          --> OK. This is equals to spr.is_deprecated = 0
483|                 --> Also empty, I don't know why.... Maybe a difference between null and empty string... That means deprecated ?? Bad news ?
1|2015-12-18
1|2015-12-23
756|2015-12-29
14|2016-01-05
1|2016-01-06
15|2016-01-09
1|2016-01-20
1|2016-01-27
5|2016-03-30
3|2016-05-03
8|2016-06-08
1|2016-06-22
4|2016-07-06
1|2016-07-07
1|2016-07-12
16|2016-07-21
17|2016-07-22
54|2016-07-26
15|2016-07-27
43|2016-08-04
3|2016-08-05
24|2016-08-08
12|2016-08-09
29|2016-08-10
20|2016-08-11
28|2016-08-12
18|2016-08-14
40|2016-08-15
36|2016-08-16
3|2016-08-19
27|2016-08-24
45|2016-08-25
24|2016-08-26
35|2016-08-29
49|2016-08-30
4|2016-09-01
8|2016-09-07
22|2016-09-08
14|2016-09-09
9|2016-09-12
1|2016-09-13
10|2016-09-14
47|2016-09-15
41|2016-09-16
8|2016-09-17
89|2016-09-19
11|2016-09-20
173|2016-09-21
12|2016-09-22
63|2016-09-23
35|2016-09-26
33|2016-09-27
31|2016-09-28
3|2016-09-29
10|2016-09-30
8|2016-10-03
24|2016-10-05
29|2016-10-06
17|2016-10-07
36|2016-10-10
21|2016-10-11
55|2016-10-12
30|2016-10-13
10|2016-10-14
46|2016-10-17
5|2016-10-18
22|2016-10-19
24|2016-10-20
40|2016-10-21
1|2016-10-24
22|2016-10-25
2|2016-10-26
5|2016-10-31
29|2016-11-01
22|2016-11-03
8|2016-11-08
382|2016-11-09
222|2016-11-10
297|2016-11-11
613|2016-11-14
5|2016-11-15
10|2016-11-16
18|2016-11-18
25|2016-11-21
24|2016-11-22
1|2016-11-28
1|2016-12-08
1|2016-12-09
31|2016-12-14
33|2016-12-15
11|2016-12-16
1|2016-12-17
16|2016-12-19
14|2016-12-20
9|2016-12-21
62|2016-12-27
10|2016-12-29
1|2017-01-01
24|2017-01-03
1|2017-01-04
13|2017-01-09
13|2017-01-10
17|2017-01-11
17|2017-01-12
3|2017-01-13
45|2017-01-17
41|2017-01-18
55|2017-01-19
18|2017-01-20
8|2017-01-23
2|2017-01-24
7|2017-01-25
3|2017-01-26
32|2017-01-27
16|2017-01-30
9|2017-01-31
35|2017-02-01
38|2017-02-02
12|2017-02-03
110|2017-02-06
163|2017-02-08
13|2017-02-09
14|2017-02-10
19|2017-02-13
10|2017-02-14
8|2017-02-15
9|2017-02-16
8|2017-02-17
1|2017-02-21
3|2017-02-22
3|2017-02-23
15|2017-02-24
4|2017-02-27
6|2017-02-28
7|2017-03-06
50|2017-03-07
20|2017-03-08
9|2017-03-09
4|2017-03-10
15|2017-03-13
4|2017-03-14
17|2017-03-15
22|2017-03-16
18|2017-03-17
27|2017-03-20
30|2017-03-21
25|2017-03-23
22|2017-03-24
1|2017-03-26
15|2017-03-27
11|2017-03-29
3|2017-03-30
13|2017-03-31
2|2017-04-03
20|2017-04-04
1|2017-04-05
2|2017-04-10
15|2017-04-24
5|2017-04-25
22|2017-04-26
13|2017-04-27
9|2017-04-28
12|2017-05-01
14|2017-05-02
22|2017-05-03
9|2017-05-04
1|2017-05-05
6|2017-05-08
13|2017-05-15
5|2017-05-17
4|2017-05-22
13|2017-05-23
6|2017-05-24
1|2017-05-25
4|2017-05-26
1|2017-05-27
2|2017-05-28
1|2017-06-01
1|2017-06-02
11|2017-06-06
17|2017-06-07
4|2017-06-08
5|2017-06-12
2|2017-06-13
8|2017-06-15
7|2017-06-19
7|2017-06-20
8|2017-06-21
4|2017-07-12
7|2017-07-19
1|2017-07-20
43|2017-07-24
7|2017-07-27
9|2017-08-01
3|2017-08-03
3|2017-08-04
14|2017-08-07
2|2017-08-08
2|2017-08-09
6|2017-08-10
2|2017-08-11
2|2017-08-14
2|2017-08-21
8|2017-08-22
1|2017-08-23
4|2017-08-24
12|2017-08-25
15|2017-08-28
11|2017-08-29
11|2017-09-01
5|2017-09-05
2|2017-09-06
6|2017-09-07
6|2017-09-11
12|2017-09-12
3|2017-09-13
2|2017-09-14
3|2017-09-15
5|2017-09-18
11|2017-09-19
21|2017-09-20
1|2017-09-26
5|2017-09-27
40|2017-09-28
8|2017-09-29
20|2017-10-02
2|2017-10-05
1|2017-10-06
11|2017-10-12
7|2017-10-13
26|2017-10-16
12|2017-10-17
26|2017-10-18
2|2017-10-19
3|2017-10-23
1|2017-10-25
29|2017-10-31
1|2017-11-02
55|2017-11-03
40|2017-11-04
11|2017-11-06
1|2017-11-07
5|2017-11-08
6|2017-11-09
6|2017-11-10
4|2017-11-14
22|2017-11-15
1|2017-11-16
1|2017-11-20
1|2017-11-22
2|2017-11-27
1|2017-11-28
3|2017-11-29
1|2017-11-30
4|2017-12-01
2|2017-12-05
1|2017-12-06
1|2017-12-07
26|2017-12-11
1|2017-12-12
3|2017-12-13
3|2017-12-14
2|2017-12-15
1|2017-12-16
73|2017-12-20
1|2017-12-22
34|2017-12-27
43|2017-12-28
61|2017-12-29
4|2017/02/21
8|2018-04-02
7|2018-04-18
14|2018-04-19
1|2018-06-04
3|2018-08-05
3163|2018-08-15
26|2018-08-17
3012|2018-08-22
3|2018-08-27
1|2018-08-28
1|2018-09-05
1|2018-10-09
13|2018-10-30
214|2018-11-08
2|2018-11-28
5|2018-12-07
1|2018-12-18
16|uuuu                  --> OK, this is equals to spr.is_deprecated = -1
```

orangejulius:
Hey @Joxit,
I haven't reviewed all the code yet, but this looks really good overall. In trying it out I've had some download issues, always with the main sqlite database. Here's one example of an error:

```
done downloading whosonfirst-data-latest.db.bz2                                      
error downloading whosonfirst-data-latest.db.bz2Error: stderr maxBuffer exceeded     
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current      
                                 Dload  Upload   Total   Spent    Left  Speed        
 18 6454M   18 1199M    0     0   474k      0  3:52:16  0:43:10  3:09:06  481k 
```

To my knowledge it wasn't caused by disk or RAM issues. The sqlite DB is bigger than any of the bundles, do you think there's a possibility of the existing download code having problems as filesizes grow?

I'll let you know how it looks after I've fully tested an import.

Joxit:
Hey,

It's not the first time I see this error, it's related to node and child process. On a low connection or big file, curl do too much output for node and fulfill the buffer.
We shoul invoke curl in silent mode and this will fix the issue (even for bundles).

orangejulius:
Hey @Joxit,
I was able to test out this code and it seems to work well!

I had to change the exec line in `./bin/start` to `exec node --max_old_space_size=4096 import.js` in order to avoid memory limit issues, but it imports just fine.

I did notice it's quite a bit slower to import than the file based import (38 minutes vs 11 minutes on my system), but that's not a huge concern, especially considering the pip-service looks like it will actually load faster.

Joxit:
Hey @orangejulius,

Yeah ! That's cool :smile: 
Change done !

Hum... That's a bit sad if SQLite import is slower than file base import.... :disappointed:

Joxit:
@orangejulius did you see any differences in the number of imported elements ?

src/components/toJSONStream.js
```diff
@@ -0,0 +1,16 @@
+const parallelTransform = require('parallel-transform');
```
missinglink: Hi @Joxit, what's the reason for using `parallel-transform` here?

I had a quick look over the code and it doesn't seem to defer non-blocking I/O like the name implies?

missinglink:
heya, sorry it's taken so long for me to get to this, I'd like to push this forward and get it merged.

@Joxit do you have any other commits you'd like to add or is it good-to-go? Do you have any experience running it in a production setting?

The whosonfirst bundles are falling quite far out of date and the time is now right to drop support for them.

missinglink:
I just tested this out on with `pelias/docker` for the `portland-metro` project and it ran for a really long time...

It seems as though the [importPlace](https://github.com/pelias/docker/blob/master/projects/portland-metro/pelias.json#L89) section is not being honoured by the `whosonfirst` importer in `sqlite: true` mode, I was importing the whole planet 😝 

I'm guessing your intention is that the user should download the full planet sqlite database and then use the `bin/sqlite-clean` script to remove any superfluous rows from the database, reducing the filesize in the process?

If that's the case then we will need to modify `bin/sqlite-clean` to honour `importPlace`.

Although I would suggest a different approach...

We could instead simply filter out the superfluous records within the query clause itself.

The advantage would be that the sqlite database remains unmodified on disk, so developers can continue using it for debugging and for later expanding their geographic target area without re-downloading the massive db again each time.

This would also have the advantage of playing well with tooling that hashes the database when comparing against a copy stored on the whosonfirst.org servers.

I think in some production use-cases it might be beneficial to 'prune' the database using `bin/sqlite-clean`, so it's great to have around for that use-case if required, but for the general populous I think having more bytes on disk is a good tradeoff when they aren't yet sure of what data they may or may not need from the planet database.

Let me know what you think @Joxit, I know you've spent a bunch of time on this already, so I could make the dev work and you could do the review if that sounds preferrable to you?

src/components/toJSONStream.js
Joxit: Hi, I thought that having the process in parallel would improve performance during the SQLite query. But you're right, it may not change anything.
I did not compare the two versions :confused:.

Joxit:
It's ok :wink:, thank you for your review !

Yes, we are using this in production since last month, especially for pip-service and it works like a charm.

Yes, I was targeting only full planet import with this PR, which need too many files. That why importPlaces is ignored :sweat_smile: 

You are right, `bin/sqlite-clean` should honour `importPlaces`, I completely forgot this part. This can be also useful for small configurations (laptop).

I think filter the records in the query will increase the import time and will be slower than the file-based version. We should try :smile: 

I will try to make the changes before I go to Japan at the end of the month.

src/components/toJSONStream.js
missinglink: I suspect this will either have no impact on performance or it'll slightly slow things down.

Node is single-threaded and so the only way to achieve 'parallelism' is to schedule tasks on to the next 'tick' of the event loop.
This only ever has a benefit for non-blocking I/O where the 'task' each cycle is simply checking if it's ready to return data, and then only using large amounts of CPU when there is something ready to do.

It's more like 'deferring work' on a single thread rather than true parallelism.

In this case, the task is `JSON.parse()` which is a CPU-bound blocking operation, so AFAIK there is no way it could benefit from any parallelism, if anything it'll slow things down due to the additional bookkeeping overhead ☹️ 

I'd be happy to be proven wrong, otherwise, I think we can just remove it and use a regular stream for simplicity sake 😸

Joxit: Yep, you were right ! I did some tests and this does not improve performance, now I use simple through2 :smile: 

Thx for this review ! :grin:

Joxit:
Hi @missinglink !

Now this PR supports `importPlaces` via sqlite import :grin:

missinglink:
Thanks @Joxit I had some issues running it today but solved them all in https://github.com/pelias/whosonfirst/pull/424, could you please review and merge those commits in here?

I've made some changes to how the iterator works based off [the docs](https://nodejs.org/api/stream.html#stream_readable_read_size_1):

> ... _read() should continue reading from the resource and pushing data until readable.push() returns false

Hopefully this results in a speed improvement, combined with increasing the `highwaterMark` will mean that `32` records will now be read from the database per tick of the event loop instead of reading 1 per tick.

missinglink:
Something is broken with the Travis-CI integration, the tests are showing as passing here https://travis-ci.org/pelias/whosonfirst/builds/506783601

missinglink:
I think this is good-to-merge now.

I like that this is disabled by default, but I'd also like to make this the new default setting once it's been tested thoroughly.
For the time being, It's important that we continue to support backwards compatibility with the bundles and continue fixing any errors that arise relating to bundles.

After merging this I'm going to test it again and then I will gradually update the projects in `pelias/docker` to use config `sqlite: true`, the docker projects are fairly active and so we'll get feedback fairly quickly if something is not working correctly.

PR https://github.com/whosonfirst/go-whosonfirst-sqlite-features/pull/4 got merged using the same index name I chose, so the good news is that soon enough the canonical distributions will include this index by default, in the interim, there will be a first-access penalty incurred when generating the index in databases which don't have it.

Overall it's a really important addition to the codebase, so thanks @Joxit for all your work 🎉

orangejulius:
Yeah, this is really awesome, and represents an important step forward in managing WOF data.

Two questions:
- Is it still possible to override the location where downloads come from via the `dataHost` config option (last one in the [list of configuration options](https://github.com/pelias/whosonfirst#configuration)? This is important for those of us running services based on WOF data as it allows us to provide our own custom data patches without having to wait for the WOF team to merge
- With all the performance increases from indicies, better stream handling, etc, what sort of loading performance do we see now? The most important metric is time to load a full set of WOF data into the PIP-service.

missinglink:
@orangejulius good questions

A. This code doesn't include any download functionality, it just scans the `$DATA_DIR/whosonfirst/sqlite` directory for `.db` files which have previously been downloaded.

B. I'm can't give a definitive number for this yet, but we will certainly have to do some testing there before enabling `sqlite: true` in production environments.

I'm going to go ahead and merge this under the feature flag which disables it by default, we can work towards our goal of switching from bundles to sqlite databases over the next few weeks as our confidence in it emboldens.

PR #423, opened 2019-02-27
index aliases for name.default

missinglink:
Index `name:eng_x_preferred` and `name:eng_x_variant` as aliases for `name.default`.

Today I was trying to search for [Koh Samui](https://spelunker.whosonfirst.org/id/102024421/) (the Thai island) but was unable to find it in Pelias.

"Koh" in Thai means island, and is sometimes also spelled "Ko", although from my experience it's less common.
Without this PR it's not possible to find any Thai islands in Pelias using the prefix "Koh".

This will also hopefully have other benefits, our current method (`wof:label OR wof:name`) of finding the name is a bit naive and probably misses out a bunch of useful aliases.

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -16,6 +16,12 @@ const population_hierarchy = [
   'ne:pop_est'
 ];
 
+// WOF fields to use for aliases
+const name_alias_fields = [
+  'name:eng_x_preferred',
+  'name:eng_x_variant'
+];
+
 // this function is used to verify that a US county QS altname is available
 function isUsCounty(base_record, wof_country, qs_a2_alt) {
   return 'US' === wof_country &&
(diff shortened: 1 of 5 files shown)


missinglink:
@orangejulius what do you think are appropriate next steps for this?

1. Merge it as-is since it only affects aliases and so likely won't have much of a damaging effect.
2. Do a planet build with regression testing before merging
3. Something else?

Please add your notes so we can track the feature and don't lose it in the weeds.

orangejulius:
Lets run a planet build with this change. Its pretty quick now.

missinglink:
:shipit:

orangejulius:
I just rebased this onto latest master, updated one test from https://github.com/pelias/whosonfirst/pull/417 which was now failing, and ran the full planet build we made last week through the  acceptance tests.

It showed no major regressions in the main suite and a small improvement in the autocomplete tests. So this looks real good!

PR #429, opened 2019-03-19
allow the sqlite_extract_data util to extract the entire contents of the database

missinglink:
The `utils/sqlite_extract_data.js` script can be used to extract the contents of a WOF sqlite database to disk.
It creates the correct directory structure, writes the geojson files and generates the meta files to emulate the format used by bundles.

Prior to this PR the script would only work when used with the `whosonfirst.importPlace` config option.
This PR adds the ability to omit `whosonfirst.importPlace` and extract the entire contents of the database to disk.


missinglink:
Tested with `whosonfirst.importPlace` unset and set, works great.

missinglink:
Build passed https://travis-ci.org/pelias/whosonfirst/builds/508346809

PR #445, opened 2019-05-16
Convert string lat/lon values to numbers

orangejulius:
While Elasticsearch ultimately does not care about whether lat/lon values are numbers or strings, other parts of code do care, specifically detecting whether records are on Null Island, and therefore likely invalid.

In particular, the work in https://github.com/pelias/whosonfirst/pull/370 assumed only the string value '0.0' would be possible for lat/lons.

By converting to numbers, we should finally be correctly filtering out the many postal codes that are on Null Island.

Connects https://github.com/pelias/pelias/issues/692

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -43,20 +43,23 @@ function getPopulation( props ) {
           find((val) => { return val >= 0; } );
 }
 
-function getLat(properties) {
-  if (properties['lbl:latitude']) {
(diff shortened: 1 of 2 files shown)


orangejulius:
Oops, I forgot that the null island checks run very early in the import pipeline (this is good for performance), so this code has no effect. https://github.com/pelias/whosonfirst/pull/447 solves the problem.

PR #447, opened 2019-05-16
fix(NullIsland): Detect string values for Null Island

orangejulius:
The stream component to filter out Null Island related records was checking against a single string value that represents 0, when of course there are many.

By converting the field into a number and then comparing with 0, Null Island detection is made more robust.

Connects https://github.com/pelias/whosonfirst/pull/370
Connects https://github.com/pelias/pelias/issues/692


orangejulius:
In my testing, we import 3998053 postalcodes before this change, and 3148499 after it.

PR #470, opened 2019-10-02
anchor WOF_NAMES_REGEX

missinglink:
anchor WOF_NAMES_REGEX to prevent matching properties such as "label:eng_x_preferred_placetype"

I noticed that we are importing the `label:eng_x_preferred_placetype` property due to a regex error.

### DIFF
--- a/src/components/extractFields.js
+++ b/src/components/extractFields.js
@@ -27,7 +27,7 @@ const NAME_ALIAS_FIELDS = [
   'label:%s_x_preferred'
 ];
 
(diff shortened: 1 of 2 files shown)


missinglink:
@Joxit FYI I just noticed this subtle bug, I think this fix is correct?

missinglink:
Hmm yes, but they are currently being indexed as aliases for `default` so it should be fine, I'm open to adding that back if it's an issue.

Issue #469, opened 2019-09-18
Do not import both `whosonfirst-data-latest.db` and `whosonfirst-data-admin-$$-latest.db`


slvlirnoff:
From this page it seems that https://dist.whosonfirst.org/sqlite/ `whosonfirst-data-latest.db` hasn't been updated since May while many `whosonfirst-data-admin-$$-latest.db` have been updated several times.

Is there some information within `whosonfirst-data-latest.db` that is not within all of `whosonfirst-data-admin-$$-latest.db`?

From https://github.com/whosonfirst-data/whosonfirst-data repository README: 
> Disclaimer
> As of May 2019, the whosonfirst-data repository has split into per-country repositories. You can read more about that change here. While we still track all issues in this repository, the data itself will live in the per-country repositories for the foreseeable future.
> 
> Per-country repositories have the following repository naming convention:
> 
> whosonfirst-data-admin-{2-char country code}

With `pelias/docker` and for the whole planet using `sqlite: true` the scripts imports both data-latest and the various data-admin per country (using inventory.json).

This seems to cause an issue for Switzerland for instance: many layers have been fixed recently (localities/localadmin https://github.com/whosonfirst-data/whosonfirst-data-admin-ch/pull/4), since the importation order of the file is:
* `whosonfirst-data-admin-$$-latest.db` (includes the fix)
* `whosonfirst-data-latest.db`. (older records, don't include the fix)

The old records are replacing the updated ones since they mostly share the same whosonfirst ids.

Related: https://github.com/pelias/whosonfirst/issues/437
Also probably related: https://github.com/pelias/pelias/issues/811

Best,
Cyprien

orangejulius:
Hi @slvlirnoff,
Thanks for diving in and investigating. There's definitely a bug in the Pelias code here. The Pelias whosonfirst importer code predates the existence of the country-specific WOF sqlite databases, and the intent of the code was that only the global database would be downloaded.

So one fix would be to fix the bug and ensure only the global sqlite db is ever downloaded. I know the WOF team is working on a process to produce up to date global sqlite db, so it will be up to date again soon. This approach would probably be the easiest and require changing the least amount of Pelias code.

The other approach would be to start using the country sqlite dbs, since they are more up to date and many people have requested a way to import data for only specific countries without having to download the whole planet.

NickStallman:
My 2c would be for option 2 since it is cleaner and more versatile and doesn't rely on WOF to complete their planet db files again.

[...]

Issue #437, opened 2019-04-17
download only whosonfirst-data-latest.db


tadjik1:
Hi there,
I'm working currently on infrastructure for pip-service and it requires sqlite dump to be downloaded first. It turns out that this module downloads more than 150 dbs from dist.whosonfirst, even though later during pip-service start only 1 main file `whosonfirst-data-latest.db` is used. Is there a way to download only 1 file contains all data and not all of them?

ref: https://github.com/pelias/wof-admin-lookup/blob/master/src/pip/readStream.js#L21

orangejulius:
Hey @tadjik1~

The [download script](https://github.com/pelias/pip-service/blob/master/bin/download) in the pip-service is configured to only download administrative data, so it should download only the file you need.

Have you given that one a try? If it downloads extra data, we want to know about it, since that's the whole point of that script.

missinglink:
Are the other 149 databases for postcodes?

If the config setting `imports.whosonfirst.importPostalcodes` is set to `true` (defaults to `false`) then I believe they will be included in the list of files to download.

Postcodes are not loaded in to PIP, although there is a long discussion about why that it the case, I won't go in to detail here.

The postcodes are imported in to elasticsearch under the `postcode` layer.

Issue #460, opened 2019-08-19
Default `whosonfirst.sqlite` to true


orangejulius:
We have been testing the SQLite download option introduced in https://github.com/pelias/whosonfirst/pull/417 for some time, and it appears to be pretty solid.

As far as I know, the SQLite databases are the preferred distribution method for WOF data going forward.

What is left before we consider setting `whosonfirst.sqlite` to true by default?

orangejulius:
There's at least one blocker for this: the wof-admin-lookup tests currently fail when `sqlite` is set to `true`: https://github.com/pelias/wof-admin-lookup/issues/262

Joxit:
Now that pelias/wof-admin-lookup#262 is fixed, I think we can set SQLite by default :smile:.

orangejulius:
We've done lots of testing with SQLite DBs recently. They are definitely the way of the future and already work well today.

Two PRs left to merge before we can do this:

- https://github.com/pelias/placeholder/pull/157 (and https://github.com/pelias/placeholder/pull/169 which improves it) to add SQLite support to Placeholder
- #477 to avoid downloading both the full planet and country DBs from dist.whosonfirst.org

Should happen soon!

PR #477, opened 2020-01-09
feat(download): Filter out country-specific SQLite DBs

orangejulius:
The country-specific SQLite DBs currently available for download from https://dist.whosonfirst.org have a few integrity issues. This causes problems in Pelias such as addresses in NYC or Berlin not having proper admin information.

The planet-wide SQLite DB, while a little out of date, is overall more valid.

While in the long term we probably want to utilize the country-specific DBs, for now it makes sense to filter them out.

Fixes https://github.com/pelias/whosonfirst/issues/469
Connects https://github.com/pelias/whosonfirst/issues/460
Fixes https://github.com/pelias/whosonfirst/issues/437


orangejulius:
We definitely don't want this PR now :)

Issue #480, opened 2020-02-23
 bzip2 to lbzip2 migration to use all CPU cores


loadit1:
Hi! 

`Line 30` in [sqlite_download.sh](https://github.com/pelias/whosonfirst/blob/master/utils/sqlite_download.sh ) : 

> bunzip2 -f "${LOCAL_BZ2_PATH}" > "${LOCAL_DB_PATH}"

`Line 53` in [download_sqlite_all.js](https://github.com/pelias/whosonfirst/blob/master/utils/download_sqlite_all.js ) : 

> extract = `bunzip2`;

Would you be so kind to advice me why not to migrate from `bzip2` to `lbzip2` in order to increase performance by using all CPU cores? 

Thanks!

missinglink:
Sounds reasonable, I'm not familiar with `lbzip2`.

The only negative I could see is that it uses more RAM but this is probably a decent trade-off considering the reduction in wall-clock decompression time.

https://vbtechsupport.com/1614/

Could you please open a PR and do some testing to ensure compatibility?

orangejulius:
Yeah, agreed. Speeding up the process of decompressing massive WOF `bzip2` archives sounds good to me :) I'd be happy to see something like this tested out.

Since I imagine `lbzip2` is not commonly installed on most systems, perhaps we can use it only if its present? We could add it to the Pelias docker images if it does show significant performance benefit.

missinglink:
Hmmm so according to that benchmark I linked `lbzip2` used over 100x as much RAM as `bzip2`.

Definitely worth investigating a little bit more and possibly detecting/logging OOM errors on modest systems.

loadit1:
Hi! 

PR is just created. 

**Summary of changes:** 
1. If `lbzip2` installed on system we use it. If not, using legacy `bzip2`.
2. Updated `Dockerfile` to install `lbzip2`
3. Additional dependency:  `command-exists`

Tested `lbzip2` version on my machine for command `npm run download`. Results are below:

**Using lbzip2**

> 542.68user 101.13system **1:49.16**elapsed 589%CPU (0avgtext+0avgdata **210392**maxresident)k
> 201176inputs+15375728outputs (1001major+1221746minor)pagefault 0swaps

**Using bzip2**

> 526.88user 168.91system **3:01.95**elapsed 382%CPU (0avgtext+avgdata **48292**maxresident)k
> 91608inputs+15375736outputs (275major+70098minor)pagefaults 0swaps

[...]

PR #483, opened 2020-02-25
support lbzip2

missinglink:
This PR contains https://github.com/pelias/whosonfirst/pull/481 with another commit to tidy up the formatting/indentation.
See https://github.com/pelias/whosonfirst/issues/480 & https://github.com/pelias/whosonfirst/pull/481 for more information.
Credit for the work goes to @loadit1 and the authorship in git has been maintained for posterity.

I've also prefixed the latest commit with `feat(xxx):` which will result in a minor version bump.

Closes https://github.com/pelias/whosonfirst/pull/481
Resolves https://github.com/pelias/whosonfirst/issues/480

### DIFF
--- a/utils/download_sqlite_all.js
+++ b/utils/download_sqlite_all.js
@@ -4,6 +4,7 @@ const fs = require('fs-extra');
 const os = require('os');
 const path = require('path');
 const downloadFileSync = require('download-file-sync');
+const commandExistsSync = require('command-exists').sync;
 
 const config = require('pelias-config').generate(require('../schema'));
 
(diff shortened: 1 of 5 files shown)


utils/download_sqlite_all.js
```diff
@@ -50,9 +51,19 @@ function download(callback) {
   const generateCommand = (sqlite, directory) => {
     let extract;
     if (/\.db\.bz2$/.test(sqlite.name_compressed)) {
-      extract = `bunzip2`;
-    } else if(/\.db\.tar\.bz2$/.test(sqlite.name_compressed)) {
-      extract = `tar -xjO`;
+      // Check if we have lbzip2 installed
+      if (commandExistsSync('lbzip2')) {
+        extract = 'lbzip';
```
orangejulius: When testing this code I ran into an issue with this command. Shouldn't it be `lbzip2` here?

utils/download_sqlite_all.js
```diff
@@ -50,9 +51,19 @@ function download(callback) {
...
+      } else {
+        extract = 'bunzip';
```
loadit1: according tar man page, flag `-j` is for `bzip2` not `bzip`. I belive that this is just typo and here we should has `bunzip2` command as it was on my original PR

orangejulius: Makes sense. I just put in #486 to fix line 56, which had a similar error. Want to open another PR? ;)

loadit1: Sure, updated in #484

PR #492, opened 2020-03-25
[i18n] Add all hierarchy names

Joxit:
## Background

With pelias/api#1300 I added some cool feature for i18n autocomplete. But there are some limitations, such as the search of `Parijs, Frankrijk`, due to hierarchy index limitation.
Since we have pelias/whosonfirst#446 and the support for multi-lang names property  in OSM, we can search for both OSM venues/addresses and WOF by their non-default/English name.

This means that if we do an autocomplete of `Parijs` (with `lang=nl` and pelias/api#1300) we will get Paris :+1: but if you do an autocomplete of `Parijs, Frankrijk` we will get no results :-1:.

## Why ?

Because the API will search a document with a name property `Parijs` AND in its hierarchy (country, locality...) `Frankrijk`. But in our ES index, we only have the default name in document hierarchy. So `Frankrijk` is not in the `Parijs` hierarchy... Only `France` is present.

## How to get rid of this ? 

We do not have many solutions... I found only 2.

1. We add alternative names in our hierarchy to takes advantage of ES fuzzy search when we autocomplete things.
2. We do 2 requests, my example is for the search of `Pont Neuf, Parijs, Frankrijk`:
    1. The first one is searching ES admin information, something like `source=wof AND (name.lang='Parijs' or name.lang='Frankrijk')` and take their ids.
    2. The second one is searching in ES `Pont Neuf` which is in the hierarchy found lately, something like `name.lang='Pont Neuf' AND (country.id IN ('whosonfirst:country:85633147', 'whosonfirst:locality:101751119') OR locality IN ('whosonfirst:country:85633147', 'whosonfirst:locality:101751119'))`

In this PR I tried the first solution and here are my results for France:

### All names in `parent.placetype`

First try with all names in `parent.placetype` without duplicates.

|   | master  | PR  |
|---|---|---|
| Index Size  |  185M | 1.1G  |
| Index Time  | 86s  | 228s  |

- Pros:
  - `Parijs, Frankrijk` works with pelias/api#1300 and `lang=nl`
- Cons:
  - The index size... 6 times bigger only with WOF data !
  - The payload is bigger when we retrieve documents because we have the full list
  - 2.6 times slower (maybe because of [this](https://github.com/pelias/model/blob/37c4330028dad260038ab05befb81f119f24a67a/Document.js#L329-L331))

This first test is not the right one in my opinion. I will try some alternatives :smile: 

related: pelias/api#1296
related: pelias/api#1300

### DIFF
--- a/src/peliasDocGenerators.js
+++ b/src/peliasDocGenerators.js
@@ -53,7 +53,15 @@ function assignField(hierarchyElement, wofDoc) {
       }
 
       break;
+    default:
+      return;
   }
+
+  _.values(hierarchyElement.name_langs).forEach(names => {
+    names.forEach(name => {
+      wofDoc.addParent(hierarchyElement.place_type, name, hierarchyElement.id.toString());
+    });
+  });
 
 }
 
--- a/test/functional.js
+++ b/test/functional.js
@@ -26,9 +26,11 @@ tape('functional', function(test) {
         t.ok(dbRecords[101772677]);
         t.deepEqual(recordOrder, ['85633147', '1108826387', '85683431', '404227861', '102068587', '404414453', '101772677']);
         t.deepEqual(dbRecords[101772677].parent, {
-          country: ['France'],
+          country: ['France', 'Frankreich','Französische Republik','French',
+                    'French Republic','Furansu','Gallía','République Française',
+                    'Francia','Frankrijk'],
           country_id: ['85633147'],
-          country_a: ['FRA'],
+          country_a: ['FRA',null],
           county: ['Roquemaure'],
           county_id: ['102068587'],
           county_a: [null],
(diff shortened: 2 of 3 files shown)


Joxit:
Ok, I did some test yesterday with a full **WOF only** planet import, this is not convincing 

| | master | PR |
| -- |-- | -- |
| Size | 3,2G | 48G |
| Time | 41m6,438s | 336m20,755s |

I did another test today, save multi-lang names in a field `${placetype}_names` but this causes an error 
```
[400] type=illegal_argument_exception, reason=Limit of total fields [1000] has been exceeded
```

> Defining too many fields in an index can lead to a mapping explosion, which can cause **out of memory errors** and **difficult situations to recover from**.
>
> Consider a situation where every new document inserted introduces new fields, such as with dynamic mapping. Each new field is added to the index mapping, which can become a problem as the mapping grows.

From [elastic documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/mapping.html#mapping-limit-settings)

Basically, this is not a good idea, so I will close this PR and find another solution

orangejulius:
Ah yeah, adding this many new names _probably_ won't work. There are also issues with field length that start happening with aliases (https://github.com/pelias/pelias/issues/862).

HOWEVER! It might be interesting for you to give it a shot on Elasticsearch 7.10: it has both faster indexing and [more space efficient indices](https://www.elastic.co/guide/en/elasticsearch/reference/current/release-highlights.html#more-space-efficient-indices). It probably won't make much of a difference but I'm sure happy to be surprised. :)

Joxit:
Yes you're right :disappointed: 

I did my tests with Elasticsearch 7.10 :sweat_smile:

PR #493, opened 2020-04-02
Typo for extraction it should be `lbunzip2`

Joxit:
Hi there,

I was testing #487 with the docker image `pelias/whosonfirst:joxit-data-geocode-earth` and I found a little bug ^^'
There is a typo, lbzip2 is for compression. However both lbzip2 and lbunzip2 work for extraction with tar command.


missinglink:
:+1:

PR #498, opened 2020-04-23
update SQLite schema and add filter to remove alt-geometries

missinglink:
A PR into [joxit/data-geocode.earth](https://github.com/pelias/whosonfirst/pull/487).

The `pelias/wof-admin-lookup` module requires this module and uses it to load geometries from SQLite databases.

I noticed a bunch of warnings when loading the PIP service using the local resolver as such:

```
warn: [wof-admin-lookup] invalid value lon=-122.726669, lat=45.580934, type=neighbourhood, values=[id=85893225, name=undefined, ]
warn: [wof-admin-lookup] invalid value lon=-122.729344, lat=45.577333, type=neighbourhood, values=[id=85893225, name=undefined, ]
warn: [wof-admin-lookup] invalid value lon=-122.728515, lat=45.578441, type=neighbourhood, values=[id=85893225, name=undefined, ]
```

It seems that although the WOF SQLite schema *supports* alt geometries, they never *published* data including them, whereas GeocodeEarth does, so we're adhering to the spec ahead of the canonical releases.

So this PR ensures that, by default, alt geometries are not loaded when using the `SQLiteStream` class.
We may wish to support them later but this at least gets any dependant codebases back to where they were.


orangejulius:
This made it into #487 but I guess I didn't do it in a way where this shows as merged. It's in there though: f8b6479

PR #500, opened 2020-04-23
Add tests for module exports

orangejulius:
This is a PR into #487 which adds some unit tests covering the shared resources exported by this module. We've been bitten many times in the past by removing code and forgetting to update `index.js`. This should help with that going forward.

It also fixes some that were just recently missed :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,6 @@
 module.exports = {
-  metadataStream: require('./src/components/metadataStream'),
   isActiveRecord: require('./src/components/isActiveRecord').create,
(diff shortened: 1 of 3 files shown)


orangejulius:
@Joxit I think your PR has become the new `master` branch :grin:

orangejulius:
If we merge this PR and #498 then your PR should be good to go. It should be a nice big new major version!

Joxit:
Ha ha yes, this was a nice catch :sweat_smile: 

Here we go ! The PR `master` is now the master :rocket: :full_moon:

PR #503, opened 2020-04-23
Deprecate `importVenues`

orangejulius:
BREAKING CHANGE: Because we do not expect Who's on First to update venue data, and we will likely never publish that data to the new data host from Geocode Earth, the `importVenues` option is now deprecated.

Connects https://github.com/pelias/whosonfirst/issues/94


orangejulius:
I think this is safe to merge, but we've already had one breaking change release today, maybe we hold off until next week?

If people report issues trying to download venues then we should merge this, but otherwise it's not urgent.

PR #518, opened 2020-10-13
GEO-46: Pelias WOF data importer improvement to include geo-shape dat…

ainemitch:
GEO-46: Pelias WOF data importer improvement to include geo-shape data into Pelias ES index - patch applied, tests updated and new task added to generate additional error log.

### DIFF
--- a/bin/start-error-log
+++ b/bin/start-error-log
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+rm -rf import_log.txt
+rm -rf error_log.txt
+
+node --max_old_space_size=6000 import.js  &> import_log.txt && cat import_log.txt | awk '$1 ~ /error/' >> error_log.txt
+
+echo -e "..............................."
+echo -n 'The total number of errors are: ' & wc -l error_log.txt
+echo -e "..............................."
+echo -n 'Total Error type: invalid_shape_exception [400] ' & grep "invalid_shape_exception" error_log.txt | wc -l
+echo -n 'Error type: invalid_shape_exception [400]: Provided shape has duplicate consecutive coordinates at: ' & grep "Provided shape has duplicate consecutive coordinates at:" error_log.txt | wc -l
+echo -n 'Error type: invalid_shape_exception [400]: Cannot determine orientation: ' & grep "Cannot determine orientation:" error_log.txt  | wc -l
+echo -n 'Error type: invalid_shape_exception [400]: Self-intersection at or near point: ' & grep "Self-intersection at or near point" error_log.txt  | wc -
+echo -e "..............................."
+echo -n 'Total Error type: mapper_parsing_exception [400]: Unable to Tessellate shape: ' & grep "Unable to Tessellate shape" error_log.txt  | wc -l
+echo -e "..............................."
+echo -n 'Total Error type: transaction error reached max retries ' & grep "transaction error reached max retries" error_log.txt | wc -l
+echo -n 'Total Error type: invalid resp from es bulk index operation: ' & grep "invalid resp from es bulk index operation" error_log.txt  | wc -l
+echo -n 'Total Error type: esclient error message=Request Entity Too Large, status=413: ' & grep "Request Entity Too Large"   error_log.txt  | wc -l
--- a/global_const.js
+++ b/global_const.js
@@ -1,0 +1,11 @@
+// Property to determine if the importer should be run against the Pelias ES index that contains the polygon geo-shape data.
+global.geo_shape_polygon=true;
+
+// Validation for constants
+if(global.geo_shape_polygon === true || global.geo_shape_polygon === false){
+  console.log('\n Running the importer against the ES index that contains the polygon geo-shape data: ' + global.geo_shape_polygon + '.');
+}else{
+  console.error('The global.geo_shape_polygon property in the global_const.js file has not been correctly configured.\n');
+  console.error('\nPlease set this value to true or false and try again.');
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -8,11 +8,14 @@
   "main": "index.js",
   "scripts": {
     "start": "./bin/start",
+    "start-debug": "node --inspect ./import.js",
+    "import": "node --max_old_space_size=8192 ./import.js",
     "download": "./bin/download",
     "test": "./bin/units",
     "lint": "jshint .",
     "validate": "npm ls",
-    "travis": "npm run test"
+    "travis": "npm run test",
+    "download-debug": "node --inspect ./utils/download_sqlite_all.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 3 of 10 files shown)


bin/start-error-log
```diff
@@ -0,0 +1,20 @@
...
+echo -n 'Error type: invalid_shape_exception [400]: Provided shape has duplicate consecutive coordinates at: ' & grep "Provided shape has duplicate consecutive coordinates at:" error_log.txt | wc -l
+echo -n 'Error type: invalid_shape_exception [400]: Cannot determine orientation: ' & grep "Cannot determine orientation:" error_log.txt  | wc -l
+echo -n 'Error type: invalid_shape_exception [400]: Self-intersection at or near point: ' & grep "Self-intersection at or near point" error_log.txt  | wc -
+echo -e "..............................."
+echo -n 'Total Error type: mapper_parsing_exception [400]: Unable to Tessellate shape: ' & grep "Unable to Tessellate shape" error_log.txt  | wc -l
+echo -e "..............................."
+echo -n 'Total Error type: transaction error reached max retries ' & grep "transaction error reached max retries" error_log.txt | wc -l
+echo -n 'Total Error type: invalid resp from es bulk index operation: ' & grep "invalid resp from es bulk index operation" error_log.txt  | wc -l
+echo -n 'Total Error type: esclient error message=Request Entity Too Large, status=413: ' & grep "Request Entity Too Large"   error_log.txt  | wc -l
```
orangejulius, line 20: This is interesting output, but I don't think this is how we'd like to implement it. Over in the [OpenAddresses importer](https://github.com/pelias/openaddresses/blob/d89f586df74ada814bbc379e9d20dca7bf4eeae3/lib/streams/documentStream.js#L38) we pass around a `stats` object that can record counts of various interesting events.

If you'd like to implement that, it would be fantastic, but let's do it in a separate PR.

package.json
```diff
@@ -8,11 +8,14 @@
   "main": "index.js",
   "scripts": {
     "start": "./bin/start",
+    "start-debug": "node --inspect ./import.js",
+    "import": "node --max_old_space_size=8192 ./import.js",
```
orangejulius, line 12: This `import` task just duplicates what is in `./bin/start`. Please remove it and if you'd like to add it, do so in a separate PR.

I think some other importers do define an `import` script so it could be interesting to insure consistency, although in general I think `npm start` is enough.

package.json
```diff
@@ -46,12 +49,14 @@
...
     "jshint": "^2.5.6",
+    "nodemon": "^2.0.4",
     "pelias-mock-logger": "^1.1.0",
     "precommit-hook": "^3.0.0",
     "proxyquire": "^2.0.0",
     "tap-dot": "^2.0.0",
     "tape": "^4.2.2",
-    "temp": "^0.9.0"
+    "temp": "^0.9.0",
+    "winston": "^3.3.3"
```
orangejulius, line 59: Please remove both of these as I don't see how they're used in this PR. Winston is already included through [pelias-logger](https://github.com/pelias/logger/), and I don't think we want to prescribe a tool like `nodemon`.

src/components/extractFields.js
```diff
@@ -188,20 +187,42 @@ function getHierarchies(id, properties) {
...
+        lon: getLon(json_object.properties),
+        bounding_box: getBoundingBox(json_object.properties),
+        population: getPopulation(json_object.properties),
+        popularity: json_object.properties['misc:photo_sum'],
+        hierarchies: getHierarchies(json_object.id, json_object.properties),
+        shape : json_object.geometry
+      };
+
+    }else{
+      record = {
```
orangejulius, line 210: It should be ok for you to just add `shape` unconditionally here.

global_const.js
```diff
@@ -0,0 +1,11 @@
...
+global.geo_shape_polygon=true;
+
+// Validation for constants
+if(global.geo_shape_polygon === true || global.geo_shape_polygon === false){
+  console.log('\n Running the importer against the ES index that contains the polygon geo-shape data: ' + global.geo_shape_polygon + '.');
+}else{
+  console.error('The global.geo_shape_polygon property in the global_const.js file has not been correctly configured.\n');
+  console.error('\nPlease set this value to true or false and try again.');
+  process.exit(1);
+}
```
orangejulius, line 11: Please remove this file completely. If you'd like to make whether or not shapes are imported configurable (a great idea), use [pelias/config](https://github.com/pelias/config) to do so.

A great example of how that works is in the downloader, where the URL base is configurable:
https://github.com/pelias/whosonfirst/blob/839f5b71e8e72ae75a362e8df7343fa1dfb0f5c7/utils/download_sqlite_all.js#L13

In this case, to fit with our other properties, you can check if `imports.whosonfirst.importShapes` is set, and default to _not_ importing shapes for now. Let me know if that makes sense or you need more examples.

src/components/extractFields.js
```diff
@@ -141,7 +141,6 @@ function getMultiLangNames(defaultName, properties) {
       return {
         key: key.substring(key.indexOf(':') + 1, key.indexOf(':') + 4), // get the iso part of the key name:iso_x_preferred
         value: properties[key]
-                .filter(name => !defaultName || defaultName.indexOf(name) < 0) // remove duplicate elements found in default name
```
orangejulius, line 144: The [pelias/model](https://github.com/pelias/model) library we use as a common system for describing Pelias documents now enforces this filter, so this code change won't have the effect you desire.

In general, we've found that duplicating a value in `name.default` and `name.${lang}` leads to problems. So we would like to ensure that those duplicates do **not** exist, while it seems like the intention of your change here is to enforce that they **do** exist.

Can you give some examples of queries that aren't behaving how you'd like that lead you to make this change? Please use our [Compare tool](https://pelias.github.io/compare/) if you can so we have a common environment to discuss.

orangejulius (changes requested):
Hi @ainemitch,
Thanks for sending this in. I've left quite a bit of feedback on this PR, I hope it's all helpful. In general we try to keep our PRs small and focused, so lets keep just changes required to import shapes in this PR.

Since there's a lot to go over, if you want to discuss this over email, video chat, etc let me know. Feel free to email us at team@pelias.io if needed.

ainemitch:
Hi Julien,

Many thanks for your comments they are very helpful.

In answer to your question on the reason for wanting the duplicate values in name.default and name.${lang}, we are actually attempting to build a reverse geo-coding application which supports multiple languages and we are using requests similar to the below for retrieving the location data.

curl --location --request GET 'localhost:9200/pelias_world_1.0/_search' --header 'Content-Type: application/json' --data-raw '{
    "query":{
        "bool": {
            "must": {
                "match_all": {}
            },
            "filter": {
                "geo_shape": {
                    "polygon": {
                        "shape":
                              {
                                  "type": "point",                             
							      "coordinates" : [5.420364, 43.291582]                        
							  },
                        "relation": "contains"
                    }
                }
            }
        }
    }
}'

What we found that with the duplicate names filtered out in some cases the translated value returned was the variant name rather than the preferred name (in a comparison with a WhosOnFirst Spelunker search, for example for the locality Marseille


Without duplicates the French translation is 13000 (variant)

 "bounding_box":"{\"min_lat\":43.1726492325,\"max_lat\":43.3910172471,\"min_lon\":5.22826669352,\"max_lon\":5.53250260623}",
               "name":{
                  "default":[
                     "Marseille",
                     "MRS"
                  ],
				 .....

				"fr":"13000",

With the duplicates included we would get both the preferred and variant names which is what we want for our application.


				"fr":[
                    "Marseille",
                    "13000"
                 ],

We will take a look at Pelias-conf and the OpenAddresses importer, thanks again for the information on those.
I will close this pull request so and raise a more targeted ones with just the shape update.
Thanks you again for taking the time to comment.

PR #522, opened 2021-05-15
update pelias/model to include popularity_bands PR

missinglink:
related: https://github.com/pelias/model/pull/138


missinglink:
this is ready to merge

the reason for the change is to bring in https://github.com/pelias/model/pull/138 (from `pelias-model@9.1.0`), however it's also going to bring in `pelias-model@9.0.0` as a result.

due to that, there are changes to the tests, I feel that this is correct but worth a second set of eyes 🙏

PR #524, opened 2021-07-20
upgrade pelias-model to 9.1.1

missinglink:
this simple PR pulls in the changes from https://github.com/pelias/model/pull/142


missinglink:
oh noes, I hope that triggered an `npm` release, I had to use an empty commit to trigger the checks again after disabling Circle/Travis and the squash commit looks to have removed the semantic-release prefix 🤦

PR #533, opened 2022-04-19
test: update expected messages count

missinglink:
two tests are currently failing with a fresh install of the node modules.
I'm assuming its related to https://github.com/pelias/logger/pull/62


missinglink:
agh no, I think I've been caught by this before, it's related to the settings in my local `~/pelias.json`.
the tests shouldn't be reading my local config, but 🤷‍♂️ until next time 😆

PR #538, opened 2023-07-14
CI: Update GitHub Actions config for uniformity

orangejulius:
This updates CI configuration for this repo with the following changes:
- the core configuration for running unit tests is moved to _test.yml so it can be shared
- a new pull request workflow is added that only runs on external forks, and runs unit tests
- Ubuntu versions are now pulled from a GitHub Actions variable (see https://github.com/pelias/pelias/issues/951)

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -1,0 +1,23 @@
+name: Unit Tests
+on: workflow_call
+jobs:
+  unit-tests:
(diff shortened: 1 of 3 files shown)


missinglink:
@orangejulius I noticed an issue today with using the `UBUNTU_VERSION` var is that CI fails for forks which don't also have it defined in Github, this unfortunately breaks the ability for 3rd party contributors to run the tests 😢

PR #539, opened 2023-09-25
Shape ingestion support

lokkju:
:wave: I have added the ability to optionally ingest shapes

actually returning shape geometry from elasticsearch depends on other changes in `schema` and `api`; those pull requests are also being created, and the entire stack can be seen here: https://github.com/lokkju/pelias-us-shape-autocomplete

---
admin region boundary display has been a long desired feature, but it should be optional as it increases index size. This changeset enables the optional ingestion of boundaries from whosonfirst data into the pelias-schema shape field.

- requires pelias/schema/pull/493 to actually store/return the shapes from elasticsearch

---
- added an `importShapes` option to the config, defaulting to false
- changed the extractFields signature to accept a wofConfig object
- added the extraction of the geometry field to the extractFields module
- in the peliasDocGenerators module, set the doc shape if a geometry field exists
- added and fixed up existing tests
---
Unit tests were added, and the full stack of changesets can be tried from the above repository.


dandeto:
I have a project where I want to import the WOF data into elastic including the shape (geometry). I was about to create a PR for this and found your PR. Hopefully the maintainers can merge this in soon. Sorry to ping, but perhaps this has fallen through the cracks? @orangejulius

orangejulius:
Hi @dandeto and anyone following,

It's really unlikely we'll merge this PR, for three reasons:
- First, we've found that geometry support in Elasticsearch is not that great. It's not as feature-filled as a real GIS tool, the quality is a bit low, the performance in terms of speed is slow for querying and VERY slow for importing
- Every field we add to the schema, even if unused, adds a small performance penalty. Since Pelias needs to be as fast as possible for quality autocomplete with lots of data, that means we have to make this field optional via config flag
- While we'd like to some extent to have the ability to turn on/and off features, every time we've done so it has introduced problems like increased complexity, broken features that are unused and therefore untested by the core team, and most significantly increased support burden. You inadvertently demonstrated this yourself over in https://github.com/pelias/docker/issues/217#issuecomment-2913364826 where you didn't mention you were using this PR. Not trying to pick on you directly, just to demonstrate it's so easy to do.

We know there's a need for Pelias to be able to serve geometries and do other geometry related things, but let's focus more on the actual need rather than this particular possible implementation, because I don't think it's the best path forward.

dandeto:
My use case is that I would rather query a database than use the in-memory pip service. But I see why increasing the complexity of the runtime config and the ES performance penalty is undesirable. Thanks for your feedback!

lokkju:
It would have been appreciated if you had responded with this comment long ago, as all of my related PRs have been ignored rather than informing me that it wasn't a direction you wanted to go.

I understand the desire not to add support burden for features your core development team has no desire to test/use, but at the time I filed these PRs, the collective set of changes were entirely configurable as an additional feature to be enabled, didn't break anything, and didn't significantly impact performance for querying nor import - especially as it was, again, a feature that would only be enabled by those who desired shape support. ES support for GIS has no bearing on the issue, as this feature just returns the geojson as a field, and is not doing any calculations on it.

If that's your final answer on the matter, feel free to close out this PR, and I'll not open a new one.

PR #543, opened 2023-12-11
move away from event-stream, replace with stream-mock

michaelkirk:

Part of https://github.com/pelias/pelias/issues/801

Technique cribbed from https://github.com/pelias/openaddresses/pull/457

I ran the tests before and after this change with a fresh npm install.


missinglink:
Thanks!

Issue #521, opened 2021-05-15
unit tests dependant on default pelias.json values


missinglink:
I noticed today that when I have my own local `~/pelias.json` file with values in `import.whosonfirst` then the tests fail.
Moving or deleting `~/pelias.json` make the tests pass again.

I didn't dig into it further but I'm guessing that having `countryCode: US` set is what's causing the issue..

```
    "whosonfirst": {
      "datapath": "/data/wof/sqlite",
      "sqlite": true,
      "importPostalcodes":false,
      "countryCode": "US",
      "importPlace": ["85633793"]
    }
```

see: https://github.com/pelias/whosonfirst/pull/520

orangejulius:
Ah yeah, I might have seen this at some point too. If not in this repo, I'm pretty sure it's possible in others.

I think the easiest way around this sort of problem is to always use an empty `pelias.json` when running the tests, like we did in https://github.com/pelias/api/pull/1431. It's not much work to set up, and it doesn't require going through all the unit tests to find which ones might possibly behave differently with non-standard `pelias.json` settings.

PR #546, opened 2024-03-13
feat(es8): support elasticsearch v8

missinglink:
this PR modifies the importer such that it works for both elasticsearch v7 & v8
in order to support v8 this PR drops support for v6.

resolves https://github.com/pelias/whosonfirst/issues/521


bin/units
```diff
@@ -2,4 +2,12 @@
 
 set -euo pipefail
 
+# ignore any user configuration in tests
```
missinglink, line 5: this was actually unrelated but it was something which had been annoying me for some time, prior to this change having `imports.whosonfirst.importPostalcodes=false` in your local `~/pelias.json` would cause the test suit to fail 🤦‍♂️

missinglink:
This is the final image compatible with elasticsearch v6:
https://hub.docker.com/layers/pelias/whosonfirst/v5.8.1/images/sha256-bb523b4b2d5502dd1082887bb3b0aa373763a073d66ba0469622d2312ece335e?context=explore

PR #548, opened 2024-11-07
update better-sqlite3 to work with node@v22

missinglink:
update better-sqlite3 to work with node@v22

package `pelias-wof-admin-lookup` depends on this module, which in turn is depended on by `pelias-csv-importer`, so fixing it here will resolve the issues on the downstream modules.

https://github.com/pelias/pelias/issues/950


orangejulius:
This can probably be rebased and merged now that #549 is in.

PR #549, opened 2024-11-27
Update CI to conform to a standard template and test all supported, non-EOL LTS Node.js releases

orangejulius:
This PR includes a bunch of commits made by a script that standardizes as much as possible our CI config across all repositories.

Along the way it also ensures we test all Node.js versions that are an LTS release, not EOL, and currently work with this repository. Any Github Actions that were out of date or used old Node.js versions (checkout and setup-node @v2) are also updated.

For the whosonfirst repository that means Node.js 18 and 20. We need to update better-sqlite3 for Node.js 22 support.

Also, the CI OS version is now hardcoded to `ubuntu-22.04`. We fooled around with an organization wide CI variable to configure that, but it broke CI in forks and doesn't really help us much, so it's now undone.

If there are any other differences in Github Actions Workflow files, they are also now removed by using a standard template.

Finally, because this repository has a Dockerfile and we just [updated the Docker baseimage to support Node.js 18](https://github.com/pelias/docker-baseimage/pull/29), there is an empty commit to trigger a new major version release.

Connects https://github.com/pelias/pelias/issues/950
Connects https://github.com/pelias/pelias/issues/951

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -16,5 +16,6 @@ jobs:
           node-version: '${{ matrix.node-version }}'
       - name: Run unit tests
         run: |
+          [[ -f ./bin/ci-setup ]] && ./bin/ci-setup
           npm install
           npm run ci
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -7,8 +7,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
-        node-version:
-          - 18.x
+        node-version: [ 18.x, 20.x ]
     steps:
       - uses: actions/checkout@v4
       - name: 'Install node.js ${{ matrix.node-version }}'
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -8,9 +8,6 @@ jobs:
         os:
           - ubuntu-22.04
         node-version:
-          - 12.x
-          - 14.x
-          - 16.x
           - 18.x
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/_test.yml
```diff
@@ -20,5 +16,6 @@ jobs:
           node-version: '${{ matrix.node-version }}'
       - name: Run unit tests
         run: |
+          [[ -f ./bin/ci-setup ]] && ./bin/ci-setup
```
missinglink, line 19: what does this do?

missinglink (approved):
LGTM

.github/workflows/_test.yml
missinglink: Would it be worth hiding this behind an `npm run ci-setup || true` script so the file name and path may be varied in each repo?

orangejulius: Yeah that's an option: I was weighing the possibilities when considering the implementation and actually this is the main things I'm looking to iterate on with feedback.

- First and foremost I'm looking to preserve consistency across Github Workflow files, having them identical across all our repos is a huge time saver when we update things.
- In the past, we've been pretty good about using `npm run` as the entrypoint for most scripts, so maintaining that consistency has some value for sure.
- On the flip side, I was actually thinking of avoiding the npm middleware layer, as it's caused problems in the past (see the signal swallowing in Docker that made us move away from `npm start` to `./bin/start`). Though there is also a [new builtin Node.js feature in v22](https://nodejs.org/en/blog/announcements/v22-release-announce#running-packagejson-scripts) that could help with that some day, or at least mean theres one less process running
- Being able to modify the path and filename, personally I don't see when we would need that, let me know if there's a particular case I'm forgetting

orangejulius: Gonna merge this for now to unblock other repos but we can change it later if we come up with a better idea

Issue #555, opened 2025-05-23
unable to locate sqlite folder when importing data


raphael10-collab:
In importing openaddresses/ data I get unable to locate sqlite folder error message:

configuration file `pelias.json` :  

    (base) root@WorldMap:~# cat pelias.json 
    {
      "interpolation": {
        "client": {
          "adapter": "http",
          "host": "http://localhost:9999"
        }
      },
      "imports": {
        "geonames": {
          "datapath": "/media/hdd/geonames",
          "import": [{
            "filename": "allCountries.txt"
          }]
        },
        "openstreetmap": {
          "datapath": "/media/hdd/osm",
          "import": [{
            "filename": "planet-latest.osm.pbf"
          }]
        },
        "openaddresses": {
          "datapath": "/media/hdd/openaddresses",
          "files": [
            "global.csv"
          ]
        },
        "polyline": {
          "datapath": "~/polyline",
          "files": [
            "road_network.polylines"
          ]
        },
        "whosonfirst": {
          "datapath": "/media/hdd/whosonfirst",
          "importPostalcodes": true
        }
      }
    }

    (base) root@WorldMap:~# ls -lah /media/hdd/
    total 44G
    drwxr-xr-x  6 root root 4.0K May 23 09:21 .
    drwxr-xr-x  3 root root 4.0K May 20 22:11 ..
    drwxr-xr-x  2 root root 4.0K May 21 12:15 geonames
    drwxr-xr-x 57 root root 4.0K May 23 09:21 openaddresses
    -rw-r--r--  1 root root  44G May 22 21:13 openaddresses.zip
    drwxr-xr-x  2 root root 4.0K May 21 10:13 osm
    drwxr-xr-x  2 root root 4.0K May 21 09:50 whosonfirst
    
    (base) root@WorldMap:~# ls -lah /media/hdd/geonames/
    total 1.7G
    drwxr-xr-x 2 root root 4.0K May 21 12:15 .
    drwxr-xr-x 6 root root 4.0K May 23 09:21 ..
    -rw-r--r-- 1 root root 1.7G May 21 03:28 allCountries.txt

    (base) root@WorldMap:~# ls /media/hdd/openaddresses
    ca  cl  co  cz  dk  es  fo  gg  gr  hr  is  jm  kr  kz  lt  lv  nc  no  pl  qa  ru  se  si  tt  ua  uy  vi  za
[...]

PR #557, opened 2025-06-25
Don't use sqlite directory if not present

arnesetzer:
:wave: I did some awesome work for the Pelias project and would love for everyone to have a look at it and provide feedback.

---
As the importer only accepts sqlite files (d8512186d93d67999166be6d9aa6e5b82805d397) since 2020 the `sqlite` directory in the whosonfirst data directory is redundant and causes some problems when the files are generated manually.

- https://github.com/pelias/whosonfirst/issues/555
- https://github.com/pelias/openstreetmap/issues/575
- https://github.com/pelias/csv-importer/issues/104
- https://github.com/pelias/csv-importer/issues/109
- https://github.com/pelias/csv-importer/issues/108
- https://github.com/pelias/openstreetmap/issues/544
- https://github.com/pelias/docker/issues/294

---
If the `sqlite` directory does not exist, the downloader will download the files to the root WOF directory. The same applies to the importer.
If the directory exists, the Downloader and Importer will use the files inside it. This ensures that older installations won't break or stop receiving updates.

---
1. Create the `sqlite` directory in the wof data directory
2. Run `npm run download` and `npm run start`. In this case the files should be downloaded into the sqlite directory and be imported as well
3. Remove the directory
4. Repeat step 2. In this case the files should be downloaded into the root directory and be imported as well.

One test was also added :)

### DIFF
--- a/test/bundleList.js
+++ b/test/bundleList.js
@@ -65,7 +65,7 @@ tape('bundlesList tests', (test) => {
         temp.cleanup((err) => {
           t.notOk(err);
           t.notOk(e);
-          t.deepEqual(bundlesList, SQLITE_EXAMPLE);
+          t.deepEqual(bundlesList.sort(), SQLITE_EXAMPLE);
           t.end();
         });
       });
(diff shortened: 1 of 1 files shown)


orangejulius (changes requested):
This looks good to me minus the debug output.

Personally I'm in favor of eventually removing the `sqlite` directory, it's not really meaningful anymore. I do like how this PR handles backwards compatibility, and eventually we can remove support for the `sqlite` directory entirely.

The only thing a little bit ugly is all the duplication of the directory checks, but I'm not sure it would be better if it was abstracted out, especially if its only needed temporarily.

@missinglink what do you think?

missinglink:
Worth noting that if you have two databases in the `whosonfirst/sqlite` directory and one in the `whosonfirst` directory that the latter would be silently ignored.

missinglink (approved):

utils/sqlite_clean.js
```diff
@@ -16,8 +17,12 @@ if (config.sqlite) {
       console.error(e);
       return;
     }
+    let sqlitePath = new Sqlite3(path.join(config.datapath));
+    if (fs.existsSync(path.join(config.datapath, 'sqlite'))) {
+      sqlitePath = new Sqlite3(path.join(config.datapath, 'sqlite'));
+    }
```
Joxit, line 23: sqlitePath should be a String instead of Sqlite3 object
 ```diff
-   let sqlitePath = new Sqlite3(path.join(config.datapath));
-    if (fs.existsSync(path.join(config.datapath, 'sqlite'))) {
-      sqlitePath = new Sqlite3(path.join(config.datapath, 'sqlite'));
-    }
+   let sqlitePath = path.join(config.datapath);
+    if (fs.existsSync(path.join(config.datapath, 'sqlite'))) {
+      sqlitePath = path.join(config.datapath, 'sqlite');
+    }
```

Joxit (changes requested):

utils/sqlite_clean.js
arnesetzer: Done

Joxit (approved):
LGTM

arnesetzer:
Is there anything preventing this PR from being merged? I would be very happy if it could be merged! :)

missinglink:
I still feel like the issue I mentioned in https://github.com/pelias/whosonfirst/pull/557#issuecomment-3013139206 will come back and bite us in the future, I like the idea suggested by Jones to resolve this:

> The other solution to avoid error is importing db from both sqlite folder and the parent?

